PR #5, opened 2019-08-06
Adding Distribute operation for KubeFed backend

mpeterson:
This patch implements the Distribute operation for the KubeFed backend.

### DIFF
--- a/pkg/federate/kubefed/federator.go
+++ b/pkg/federate/kubefed/federator.go
@@ -6,6 +6,8 @@ import (
 	"sync"
 
 	"github.com/submariner-io/admiral/pkg/federate"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -33,6 +35,7 @@ type federator struct {
 	startInformerOnce      sync.Once
 	clusterWatchers        []*clusterWatcher
 	stopChan               <-chan struct{}
+	scheme                 *runtime.Scheme
 }
 
 // New creates a new Federator instance.
@@ -46,6 +49,7 @@ func New(kubeFedConfig *rest.Config, stopChan <-chan struct{}) (federate.Federat
 		clusterMap:    make(map[string]*rest.Config),
 		kubeFedClient: kubeFedClient,
 		stopChan:      stopChan,
+		scheme:        scheme.Scheme,
 	}
 
 	listerWatcher, err := newKubeFedClusterListerWatcher(kubeFedConfig)
--- a/pkg/federate/kubefed/resource.go
+++ b/pkg/federate/kubefed/resource.go
@@ -1,17 +1,114 @@
 package kubefed
 
 import (
+	"context"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/klog"
+	ctlutil "sigs.k8s.io/kubefed/pkg/controller/util"
+	kubefedopt "sigs.k8s.io/kubefed/pkg/kubefedctl/options"
 )
 
 const (
-//	federatedKindPrefix string = "Federated"
+	federatedKindPrefix         = "Federated"
+	errorUnstructuredConversion = "Error converting to unstructured.Unstructured: %v"
+	errorSettingFederatedFields = "Error setting fields necessary for federating: %v"
+	errorUpdateResource         = "Error updating the resource, assuming it does not exist and creating new one: %v"
+	errorCreateResource         = "Error creating the resource: %v"
 )
 
+var systemMetadataFields = []string{"selfLink", "uid", "resourceVersion", "generation", "creationTimestamp", "deletionTimestamp", "deletionGracePeriodSeconds"}
+
 func (f *federator) Distribute(resource runtime.Object, clusterNames ...string) error {
-	panic("not implemented")
+	fedResource, err := f.createFederatedResource(resource)
+	if err != nil {
+		return err
+	}
+
+	if len(clusterNames) > 0 {
+		unstructured.RemoveNestedField(fedResource.Object, ctlutil.SpecField, ctlutil.PlacementField)
+		err = ctlutil.SetClusterNames(fedResource, clusterNames)
+		if err != nil {
+			// Path not covered by unit tests because it's trivial behavior yet complex to test
+			return fmt.Errorf(errorSettingFederatedFields, err)
+		}
+	}
+
+	// Update first, as it will probably be the most common operation
+	err = f.kubeFedClient.Update(context.TODO(), fedResource)
+	if err == nil {
+		return nil
+	}
+
+	klog.Infof(errorUpdateResource, err)
+
+	err = f.kubeFedClient.Create(context.TODO(), fedResource)
+	if err != nil {
+		// Path not covered by unit tests because it's trivial behavior yet complex to test
+		return fmt.Errorf(errorCreateResource, err)
+	}
+
+	return nil
 }
 
 func (f *federator) Delete(resource runtime.Object) error {
 	panic("not implemented")
+}
+
+func (f *federator) createFederatedResource(resource runtime.Object) (*unstructured.Unstructured, error) {
+	targetResource := &unstructured.Unstructured{}
+	err := f.scheme.Convert(resource, targetResource, nil)
+	if err != nil {
+		return nil, fmt.Errorf(errorUnstructuredConversion, err)
+	}
+
+	fedResource := &unstructured.Unstructured{}
+	setBasicMetaFields(fedResource, targetResource)
+
+	// TODO(mpeterson): Figure out how to use the one from KubeFed
+	// kubefedfed.RemoveUnwantedFields(targetResource)
+	removeUnwantedFields(targetResource)
+
+	err = unstructured.SetNestedField(fedResource.Object, targetResource.Object, ctlutil.SpecField, ctlutil.TemplateField)
+	if err != nil {
+		// Path not covered by unit tests because it's trivial behavior yet complex to test
+		return nil, fmt.Errorf(errorSettingFederatedFields, err)
+	}
+	err = unstructured.SetNestedStringMap(fedResource.Object, map[string]string{}, ctlutil.SpecField, ctlutil.PlacementField, ctlutil.ClusterSelectorField, ctlutil.MatchLabelsField)
+	if err != nil {
+		// Path not covered by unit tests because it's trivial behavior yet complex to test
+		return nil, fmt.Errorf(errorSettingFederatedFields, err)
+	}
+
+	return fedResource, nil
+}
+
+func setBasicMetaFields(resource, base *unstructured.Unstructured) {
+	fedKind := federatedKindPrefix + base.GetKind()
+	resource.SetKind(fedKind)
+
+	gv := schema.GroupVersion{
+		Group:   kubefedopt.DefaultFederatedGroup,
+		Version: kubefedopt.DefaultFederatedVersion,
+	}
+	resource.SetAPIVersion(gv.String())
+
+	resource.SetName(base.GetName())
+	resource.SetNamespace(base.GetNamespace())
+}
+
+func removeUnwantedFields(resource *unstructured.Unstructured) {
+	for _, field := range systemMetadataFields {
+		unstructured.RemoveNestedField(resource.Object, "metadata", field)
+		// For resources with pod template subresource (jobs, deployments, replicasets)
+		unstructured.RemoveNestedField(resource.Object, "spec", "template", "metadata", field)
+	}
+	unstructured.RemoveNestedField(resource.Object, "metadata", "name")
+	unstructured.RemoveNestedField(resource.Object, "metadata", "namespace")
+	unstructured.RemoveNestedField(resource.Object, "apiVersion")
+	unstructured.RemoveNestedField(resource.Object, "kind")
+	unstructured.RemoveNestedField(resource.Object, "status")
 }
--- a/pkg/federate/kubefed/resource.go
+++ b/pkg/federate/kubefed/resource.go
@@ -71,10 +71,13 @@ func createFederatedResource(scheme *runtime.Scheme, resource runtime.Object) (*
 	// kubefedfed.RemoveUnwantedFields(targetResource)
 	removeUnwantedFields(targetResource)
 
+	// Adding the template field and including the targetResource's object as its content.
 	err = unstructured.SetNestedField(fedResource.Object, targetResource.Object, ctlutil.SpecField, ctlutil.TemplateField)
 	if err != nil {
 		return nil, fmt.Errorf(errorSettingFederatedFields, err)
 	}
+	// Adding an empty selector for the Placement field's MatchLabel. An empty selector means
+	// all clusters.
 	err = unstructured.SetNestedStringMap(fedResource.Object, map[string]string{}, ctlutil.SpecField, ctlutil.PlacementField, ctlutil.ClusterSelectorField, ctlutil.MatchLabelsField)
 	if err != nil {
 		return nil, fmt.Errorf(errorSettingFederatedFields, err)
@@ -98,6 +101,10 @@ func setBasicMetaFields(resource, base *unstructured.Unstructured) {
 	resource.SetNamespace(base.GetNamespace())
 }
 
+// This function removes metadata information that is being added by a running Kubernetes API
+// and that effectively exists only on objects retrieved from the API. This metadata is thus
+// only of interest for the objects living in the API but not for a new object that we will
+// create such in our use case.
 func removeUnwantedFields(resource *unstructured.Unstructured) {
 	for _, field := range systemMetadataFields {
 		unstructured.RemoveNestedField(resource.Object, "metadata", field)
--- a/pkg/federate/kubefed/resource.go
+++ b/pkg/federate/kubefed/resource.go
@@ -4,26 +4,30 @@ import (
 	"context"
 	"fmt"
 
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/klog"
 	ctlutil "sigs.k8s.io/kubefed/pkg/controller/util"
 	kubefedopt "sigs.k8s.io/kubefed/pkg/kubefedctl/options"
 )
 
 const (
 	federatedKindPrefix         = "Federated"
-	errorUnstructuredConversion = "Error converting to unstructured.Unstructured: %v"
-	errorSettingFederatedFields = "Error setting fields necessary for federating: %v"
-	errorUpdateResource         = "Error updating the resource, assuming it does not exist and creating new one: %v"
-	errorCreateResource         = "Error creating the resource: %v"
+	errorSettingFederatedFields = "error setting fields necessary for federating: %v"
 )
 
 var systemMetadataFields = []string{"selfLink", "uid", "resourceVersion", "generation", "creationTimestamp", "deletionTimestamp", "deletionGracePeriodSeconds"}
 
+type UnstructuredConversionError struct {
+	err      error
+	resource runtime.Object
+}
+
+var _ error = &UnstructuredConversionError{}
+
 func (f *federator) Distribute(resource runtime.Object, clusterNames ...string) error {
-	fedResource, err := f.createFederatedResource(resource)
+	fedResource, err := createFederatedResource(f.scheme, resource)
 	if err != nil {
 		return err
 	}
@@ -32,7 +36,6 @@ func (f *federator) Distribute(resource runtime.Object, clusterNames ...string) 
 		unstructured.RemoveNestedField(fedResource.Object, ctlutil.SpecField, ctlutil.PlacementField)
 		err = ctlutil.SetClusterNames(fedResource, clusterNames)
 		if err != nil {
-			// Path not covered by unit tests because it's trivial behavior yet complex to test
 			return fmt.Errorf(errorSettingFederatedFields, err)
 		}
 	}
@@ -43,26 +46,22 @@ func (f *federator) Distribute(resource runtime.Object, clusterNames ...string) 
 		return nil
 	}
 
-	klog.Infof(errorUpdateResource, err)
-
-	err = f.kubeFedClient.Create(context.TODO(), fedResource)
-	if err != nil {
-		// Path not covered by unit tests because it's trivial behavior yet complex to test
-		return fmt.Errorf(errorCreateResource, err)
+	if !errors.IsNotFound(err) {
+		return err
 	}
 
-	return nil
+	return f.kubeFedClient.Create(context.TODO(), fedResource)
 }
 
 func (f *federator) Delete(resource runtime.Object) error {
 	panic("not implemented")
 }
 
-func (f *federator) createFederatedResource(resource runtime.Object) (*unstructured.Unstructured, error) {
+func createFederatedResource(scheme *runtime.Scheme, resource runtime.Object) (*unstructured.Unstructured, error) {
 	targetResource := &unstructured.Unstructured{}
-	err := f.scheme.Convert(resource, targetResource, nil)
+	err := scheme.Convert(resource, targetResource, nil)
 	if err != nil {
-		return nil, fmt.Errorf(errorUnstructuredConversion, err)
+		return nil, &UnstructuredConversionError{err, resource}
 	}
 
 	fedResource := &unstructured.Unstructured{}
@@ -74,12 +73,10 @@ func (f *federator) createFederatedResource(resource runtime.Object) (*unstructu
 
 	err = unstructured.SetNestedField(fedResource.Object, targetResource.Object, ctlutil.SpecField, ctlutil.TemplateField)
 	if err != nil {
-		// Path not covered by unit tests because it's trivial behavior yet complex to test
 		return nil, fmt.Errorf(errorSettingFederatedFields, err)
 	}
 	err = unstructured.SetNestedStringMap(fedResource.Object, map[string]string{}, ctlutil.SpecField, ctlutil.PlacementField, ctlutil.ClusterSelectorField, ctlutil.MatchLabelsField)
 	if err != nil {
-		// Path not covered by unit tests because it's trivial behavior yet complex to test
 		return nil, fmt.Errorf(errorSettingFederatedFields, err)
 	}
 
@@ -90,6 +87,7 @@ func setBasicMetaFields(resource, base *unstructured.Unstructured) {
 	fedKind := federatedKindPrefix + base.GetKind()
 	resource.SetKind(fedKind)
 
+	// TODO(mpeterson): Modify to get the type config and generalize this usage
 	gv := schema.GroupVersion{
 		Group:   kubefedopt.DefaultFederatedGroup,
 		Version: kubefedopt.DefaultFederatedVersion,
@@ -111,4 +109,12 @@ func removeUnwantedFields(resource *unstructured.Unstructured) {
 	unstructured.RemoveNestedField(resource.Object, "apiVersion")
 	unstructured.RemoveNestedField(resource.Object, "kind")
 	unstructured.RemoveNestedField(resource.Object, "status")
+}
+
+func (err *UnstructuredConversionError) Error() string {
+	return fmt.Sprintf(
+		"error converting to unstructured.Unstructured: %v\n(RES=%#v)",
+		err.err,
+		err.resource,
+	)
 }
--- a/pkg/federate/kubefed/resource_test.go
+++ b/pkg/federate/kubefed/resource_test.go
@@ -2,12 +2,12 @@ package kubefed
 
 import (
 	"context"
+	"errors"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -17,6 +17,34 @@ import (
 	kubefedopt "sigs.k8s.io/kubefed/pkg/kubefedctl/options"
 )
 
+type fakeClientWithUpdateError struct {
+	msg string
+}
+
+var _ client.Client = &fakeClientWithUpdateError{}
+
+func (fake *fakeClientWithUpdateError) Update(ctx context.Context, obj runtime.Object) error {
+	return errors.New(fake.msg)
+}
+
+func (fake *fakeClientWithUpdateError) Get(ctx context.Context, key client.ObjectKey, obj runtime.Object) error {
+	return nil
+}
+
+func (fake *fakeClientWithUpdateError) List(ctx context.Context, opts *client.ListOptions, list runtime.Object) error {
+	return nil
+}
+
+func (fake *fakeClientWithUpdateError) Create(ctx context.Context, obj runtime.Object) error {
+	return nil
+}
+
+func (fake *fakeClientWithUpdateError) Delete(ctx context.Context, obj runtime.Object, opts ...client.DeleteOptionFunc) error {
+	return nil
+}
+
+func (fake *fakeClientWithUpdateError) Status() client.StatusWriter { return nil }
+
 var _ = Describe("Federator", func() {
 	Describe("function Distribute", testDistribute)
 	Describe("helper function createFederatedResource", testCreateFederatedResource)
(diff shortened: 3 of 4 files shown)


mpeterson:
Please review "Adding Distribute operation for KubeFed backend"

"Change Distribute interface to have optional arguments" is in #4

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,308 @@
...
+
+			It("should return a null expected", func() {
+				Expect(expectedTyped).To(BeNil())
+			})
+		})
+	})
+
+	When("the resource type scheme has been added to the type registry", func() {
+		BeforeEach(func() {
+			corev1.AddToScheme(scheme)
```
mangelajo: (no action required, just different opinion)
Again I think it's more clear to the reader where does expected/err come from exactly without the need to scroll.

IMO it's better to have a bit more code, but make the tests easier to read/explicit, in contrast to more compact code.
Of course, it's a trade off, and just my opinion, not sure what others think.

pkg/federate/kubefed/resource.go
```diff
@@ -1,17 +1,114 @@
...
+		}
+	}
+
+	// Update first, as it will probably be the most common operation
+	err = f.kubeFedClient.Update(context.TODO(), fedResource)
+	if err == nil {
+		return nil
+	}
+
+	klog.Infof(errorUpdateResource, err)
```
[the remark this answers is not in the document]
mpeterson: will do

pkg/federate/kubefed/resource.go
```diff
@@ -1,17 +1,114 @@
...
+	return fedResource, nil
+}
+
+func setBasicMetaFields(resource, base *unstructured.Unstructured) {
+	fedKind := federatedKindPrefix + base.GetKind()
+	resource.SetKind(fedKind)
+
+	gv := schema.GroupVersion{
+		Group:   kubefedopt.DefaultFederatedGroup,
+		Version: kubefedopt.DefaultFederatedVersion,
```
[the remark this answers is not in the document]
mpeterson: Yeah, actually the most correct way to do this is to receive the TypeConfig just in case the user defined other API versions that are not the default (it is not going to be our case that's why it makes a lot of sense to use the defaults here)

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,308 @@
...
+		scheme = runtime.NewScheme()
+	})
+
+	JustBeforeEach(func() {
+		f = newFederatorWithScheme(scheme, initObjs...)
+		expectedTyped, err = f.createFederatedResource(resource)
+		expected = expectedTyped
+	})
+
+	When("the resource type scheme has NOT been added to the type registry", func() {
```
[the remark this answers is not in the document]
mpeterson: a When is just a Context that writes a When at the beginning of the sentence.

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,308 @@
...
+		It("should throw an error", func() {
+			Expect(err).To(HaveOccurred())
+		})
+	})
+
+	When("the resource type scheme has been added to the type registry", func() {
+		BeforeEach(func() {
+			corev1.AddToScheme(scheme)
+			tmpFederator := newFederatorWithScheme(scheme)
```
mkolesnik: I find it weird that you need another federator here, since it'll also use another client with another cache, but you only use it to create the "federated resource" and then dump it..

IMHO it's confusing and a bit counter-intuitive.

Might I suggest that you make createFederatedResource and independent method that receives the scheme as a parameter, which should help simplify this situation (and also the tests for createFederatedResource would be a bit simpler).
I tested it locally and it's super easy, barely an inconvenience ;)

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,308 @@
...
+		JustBeforeEach(func() {
+			fedPod = getResourceFromAPI(f.kubeFedClient, fedResource)
+		})
+
+		It("should return no error on success", func() {
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		When("the resource is already in the Kube API", func() {
+			expectedImage := "apache"
```
mkolesnik: nit: https://i.imgflip.com/37frvg.jpg

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,308 @@
...
+				fedPodObj, _, _ := getTemplateField(fedPod)
+				expectObj, _, _ := getTemplateField(fedResource)
+
+				Expect(fedPodObj["spec"]).ToNot(BeEquivalentTo(expectObj["spec"]))
+
+				mapObj, _, _ := unstructured.NestedSlice(
+					fedPod.Object,
+					ctlutil.SpecField, ctlutil.TemplateField, "spec", "containers",
+				)
+				image := mapObj[0].(map[string]interface{})["image"]
```
mkolesnik: nit: The level of drill down is quite crazy IMHO, I wonder if this can be simplified (e.g. if you compare the spec to the expected one)?

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,308 @@
...
+				Expect(image).To(BeEquivalentTo(expectedImage))
+			})
+		})
+
+		When("the resource is not in the Kube API", func() {
```
mkolesnik: Technically it would be more correct to assert that the resource is not in the client before the call to Distribute (or that there arent any resources).

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,308 @@
...
+		})
+
+		When("the resource needs to be distributed to specific clusters", func() {
+			BeforeEach(func() {
+				clusterNames = []string{"cluster1", "cluster2"}
+			})
+
+			It("should set the placement to be an explicit list of clusters", func() {
+				mapObj, found, err := getClustersField(fedPod)
+				fedClusterNames := unpackClusterNames(mapObj)
```
mkolesnik: nit: Why isn't this inline?

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,308 @@
...
+		})
+
+		When("the resource is not in the Kube API", func() {
+			It("should create the resource", func() {
+				Expect(fedPod).ToNot(BeNil())
+				Expect(fedPod).To(BeEquivalentTo(fedResource))
+			})
+		})
+
+		When("the resource needs to be distributed to specific clusters", func() {
```
mkolesnik: BTW I didn't see any check for when you don't send any cluster names, or did I miss something?

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,308 @@
...
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				corev1.Container{
+					Image: "nginx",
+					Name:  "httpd",
+				},
+			},
+		},
+	}
+	return resourceTyped
```
mkolesnik: nit: Why not simply do the return of the instance? Why do you need a named variable?

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,308 @@
...
+
+func newFederatorWithScheme(scheme *runtime.Scheme, initObjs ...runtime.Object) *federator {
+	f := &federator{
+		scheme: scheme,
+		kubeFedClient: fake.NewFakeClientWithScheme(
+			scheme,
+			initObjs...,
+		),
+	}
+	return f
```
mkolesnik: nit: Why not simply do the return of the instance? Why do you need a named variable?

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,308 @@
...
+	for _, item := range object {
+		clusterDict := item.(map[string]interface{})
+		clusterName := clusterDict[ctlutil.NameField].(string)
+		clusterNames = append(clusterNames, clusterName)
+	}
+	return clusterNames
+}
+
+func getResourceFromAPI(kubeFedClient client.Client, resource *unstructured.Unstructured) *unstructured.Unstructured {
+	fedPod := &unstructured.Unstructured{}
```
mkolesnik: s/fedPod/fedResource/?

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,308 @@
...
+	fedPod.SetGroupVersionKind(schema.GroupVersionKind{
```
mkolesnik: Can't you simply send "resource.GroupVersionKind()"?

pkg/federate/kubefed/resource.go
```diff
@@ -1,17 +1,114 @@
...
+	err = f.kubeFedClient.Update(context.TODO(), fedResource)
+	if err == nil {
+		return nil
+	}
+
+	klog.Infof(errorUpdateResource, err)
+
+	err = f.kubeFedClient.Create(context.TODO(), fedResource)
+	if err != nil {
+		// Path not covered by unit tests because it's trivial behavior yet complex to test
```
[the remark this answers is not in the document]
mpeterson: As we discussed in person it is not possible, as it wouldn't reach this place.

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,308 @@
...
+
+		It("should return no error on success", func() {
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		When("the resource is already in the Kube API", func() {
+			expectedImage := "apache"
+
+			BeforeEach(func() {
+				initObjs = append(initObjs, fedResource)
```
mkolesnik: Instead of exposing internals here, you can have the tests use Distribute to create the original, and just test that a second call to Distribute actually updates it.

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,308 @@
...
+
+				resourceTyped.Spec = corev1.PodSpec{
+					Containers: []corev1.Container{
+						corev1.Container{
+							Image: expectedImage,
+							Name:  "httpd",
+						},
+					},
+				}
+			})
```
tpantelis: I'm seeing unexpected behavior wrt closure vars that's causing unexpected test results.  The way you've structured your nested `BeforeEach` and `JustBeforeEach ` blocks, you're assuming that, eg, the affects of the `BeforeEach` on line 170 are not observed by the `It` on line 199 since they don't share their immediate `When` parent (I assumed the same). However this is not necessarily the case. If this block runs before line 199 then it leaves `initObjs` set. To alleviate this leakage, you would have to add an `AfterEach` above that resets `initObjs` et al. However I think these side effects would make the tests harder to understand and maintain - I go back to my previous comment of eliminating/reducing the use of closures and  eliminating the use of nested `BeforeEach` and `JustBeforeEach ` blocks altogether.

pkg/federate/kubefed/federator.go
```diff
@@ -46,6 +49,7 @@ func New(kubeFedConfig *rest.Config, stopChan <-chan struct{}) (federate.Federat
 		clusterMap:    make(map[string]*rest.Config),
 		kubeFedClient: kubeFedClient,
 		stopChan:      stopChan,
+		scheme:        scheme.Scheme,
```
tpantelis: I'm curious why you store the `scheme` as a field instead of inlining `scheme.Scheme`. The UTs set it to `runtime.NewScheme()` which `scheme.Scheme` equates to. Or am I missing something?

pkg/federate/kubefed/resource.go
```diff
@@ -1,17 +1,114 @@
...
+	"k8s.io/klog"
+	ctlutil "sigs.k8s.io/kubefed/pkg/controller/util"
+	kubefedopt "sigs.k8s.io/kubefed/pkg/kubefedctl/options"
 )
 
 const (
-//	federatedKindPrefix string = "Federated"
+	federatedKindPrefix         = "Federated"
+	errorUnstructuredConversion = "Error converting to unstructured.Unstructured: %v"
+	errorSettingFederatedFields = "Error setting fields necessary for federating: %v"
```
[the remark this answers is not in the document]
mangelajo: I agree with Tom here, I thought of it, but didn't want to nitpick.

Consts are useful on the testing side to get help from compile-time checks while using strings, or for strings of very repeated use. But... I don't see the need here. Only makes the reader jump around while reading code ;)

pkg/federate/kubefed/resource.go
```diff
@@ -1,17 +1,114 @@
...
+
+	klog.Infof(errorUpdateResource, err)
+
+	err = f.kubeFedClient.Create(context.TODO(), fedResource)
+	if err != nil {
+		// Path not covered by unit tests because it's trivial behavior yet complex to test
+		return fmt.Errorf(errorCreateResource, err)
+	}
+
+	return nil
```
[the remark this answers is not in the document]
mangelajo: :+1:

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,308 @@
...
+	})
+
+	When("the resource type scheme has NOT been added to the type registry", func() {
+		When("failing to convert to Unstructured", func() {
+			It("should throw an error", func() {
+				Expect(err).To(MatchError(HavePrefix(errorUnstructuredConversion, "")))
+			})
+
+			It("should return a null expected", func() {
+				Expect(expectedTyped).To(BeNil())
```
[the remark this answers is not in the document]
mpeterson: > Although the linter will probably warn you about not handling the return err, you could still add

IMHO a well configured linter should not apply that check for test files

pkg/federate/kubefed/resource.go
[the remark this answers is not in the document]
mpeterson: Because I prefer keep the logging on us and not depend on the consumer. I can change this if other people agree.

pkg/federate/kubefed/resource_test.go
mpeterson: I think it's clearer as is

pkg/federate/kubefed/resource_test.go
mpeterson: it was an implicit test, making it explicit now

pkg/federate/kubefed/resource_test.go
mpeterson: Because the test is for testing Distribute and I find it logically incorrect to do so.

pkg/federate/kubefed/resource.go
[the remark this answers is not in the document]
tpantelis: I think it's OK to transform the returned error message with more context as you're doing here.

pkg/federate/kubefed/federator.go
mpeterson: IIRC all types are registered in this scheme through their respective init()

tpantelis: All k8s types are. But, eg, submariner CRD types (ie Endpoint and Cluster) define their own [Scheme](https://github.com/submariner-io/submariner/blob/master/pkg/client/clientset/versioned/scheme/register.go). I ran into the issue of using the correct `Scheme` in the cluster code for watching `KubeFedCluster` - I had to use the `sigs.k8s.io/kubefed/pkg/client/generic/scheme.Scheme`. I suspect we'll run into the same issue here but don't worry about it right now - I'll find out soon enough when I start using this for the submariner CRDs 😄

pkg/federate/kubefed/resource.go
```diff
@@ -1,17 +1,114 @@
...
+	setBasicMetaFields(fedResource, targetResource)
+
+	// TODO(mpeterson): Figure out how to use the one from KubeFed
+	// kubefedfed.RemoveUnwantedFields(targetResource)
+	removeUnwantedFields(targetResource)
+
+	err = unstructured.SetNestedField(fedResource.Object, targetResource.Object, ctlutil.SpecField, ctlutil.TemplateField)
+	if err != nil {
+		// Path not covered by unit tests because it's trivial behavior yet complex to test
+		return nil, fmt.Errorf(errorSettingFederatedFields, err)
```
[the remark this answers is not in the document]
mpeterson: it's redundant, since the error returned already has that information

pkg/federate/kubefed/resource.go
```diff
@@ -1,17 +1,114 @@
...
+	if err != nil {
+		return nil, fmt.Errorf(errorUnstructuredConversion, err)
+	}
+
+	fedResource := &unstructured.Unstructured{}
+	setBasicMetaFields(fedResource, targetResource)
+
+	// TODO(mpeterson): Figure out how to use the one from KubeFed
+	// kubefedfed.RemoveUnwantedFields(targetResource)
+	removeUnwantedFields(targetResource)
```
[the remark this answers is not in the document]
mpeterson: Not sure I understand what you mean

mpeterson:
This patch should be squashed when ready to merge!

pkg/federate/kubefed/resource.go
```diff
@@ -1,17 +1,120 @@
...
 
 const (
-//	federatedKindPrefix string = "Federated"
+	federatedKindPrefix         = "Federated"
+	errorSettingFederatedFields = "error setting fields necessary for federating: %v"
 )
 
+var systemMetadataFields = []string{"selfLink", "uid", "resourceVersion", "generation", "creationTimestamp", "deletionTimestamp", "deletionGracePeriodSeconds"}
+
+type UnstructuredConversionError struct {
```
tpantelis: Do you intend this to be exported?

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,389 @@
...
+			It("should set the placement to be an explicit list of clusters", func() {
+				// This test implicitly tests that the nested hierarchy of fields exists
+				_ = f.Distribute(resource, clusterNames...)
+				fedPod, err := getResourceFromAPI(f.kubeFedClient, fedResource)
+				Expect(err).ToNot(HaveOccurred())
+
+				mapObj, found, err := getClustersField(fedPod)
+				fedClusterNames := unpackClusterNames(mapObj)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(found).To(BeTrue(), "Hierarchy of fields missing!")
```
tpantelis: Can we change this to _"The cluster field is missing!"_?

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,389 @@
...
+			})
+
+			It("should set the placement to be an explicit list of clusters", func() {
+				// This test implicitly tests that the nested hierarchy of fields exists
+				_ = f.Distribute(resource, clusterNames...)
+				fedPod, err := getResourceFromAPI(f.kubeFedClient, fedResource)
+				Expect(err).ToNot(HaveOccurred())
+
+				mapObj, found, err := getClustersField(fedPod)
+				fedClusterNames := unpackClusterNames(mapObj)
```
tpantelis: bit of a nit but this line should really be after 279 as it technically shouldn't try to use `mapObj` if an error or `!found` occurred. If `mapObj` is nil, `unpackClusterNames` does happen to handle that and not blow up.

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,389 @@
...
+
+			It("the returned resource's placement should not contain an empty matcher", func() {
+				// This test implicitly tests that the nested hierarchy of fields exists
+				_ = f.Distribute(resource, clusterNames...)
+				fedPod, err := getResourceFromAPI(f.kubeFedClient, fedResource)
+				Expect(err).ToNot(HaveOccurred())
+
+				mapObj, found, err := getMatchLabelsField(fedPod)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(found).ToNot(BeTrue(), "Hierarchy of fields is present!")
```
tpantelis: Can we change this to _"The matchLabels field is missing!"_? Also at the other call sites.  

We could  go further and have `getMatchLabelsField` do the checking that's on lines 290-291 and have it return only a valid `mapObj` - that would simplify the call sites. Same with `getClustersField` and `getTemplateField`.

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,389 @@
...
+
+			mapObj, found, err := getTemplateField(expected)
+			targetResource := &unstructured.Unstructured{}
+
+			convErr := scheme.Convert(resource, targetResource, nil)
+			Expect(convErr).ToNot(HaveOccurred())
+			removeUnwantedFields(targetResource)
+
+			Expect(err).ToNot(HaveOccurred())
+			Expect(found).To(BeTrue(), "Hierarchy of fields missing!")
```
tpantelis: Can we change this to _"The template field is missing!"_?

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,389 @@
...
+				resource = newPod("test-pod", expectedImage)
+			})
+
+			It("should update the resource", func() {
+				_ = f.Distribute(resource, clusterNames...)
+				fedPod, err := getResourceFromAPI(f.kubeFedClient, fedResource)
+				Expect(err).ToNot(HaveOccurred())
+				fedPodObj, _, err := getTemplateField(fedPod)
+				Expect(err).ToNot(HaveOccurred())
+				expectObj, _, err := getTemplateField(fedResource)
```
tpantelis: How come you're ignoring the `found` return flag here and on line 224?

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,389 @@
...
+			expected, _ = createFederatedResource(scheme, resource)
+			Expect(expected.GetName()).
+				To(Equal(resource.GetName()))
+		})
+
+		It("the returned resource's template should be the input's unstructured", func() {
+			// This test implicitly tests that the nested hierarchy of fields exists
+			expected, _ = createFederatedResource(scheme, resource)
+
+			mapObj, found, err := getTemplateField(expected)
```
tpantelis: nit: this line could be moved before its use on line 139.

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,389 @@
...
+func testDistribute() {
+	var (
+		resource      *corev1.Pod
+		fedResource   *unstructured.Unstructured
+		f             *federator
+		initObjs      []runtime.Object
+		clusterNames  []string
+		scheme        *runtime.Scheme
+		clientPatcher client.Client
+	)
```
tpantelis: There's more we can do here to simplify and reduce the use/scope of closures but I can look at it later and submit a separate PR. No need to hold this one up for that.

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,389 @@
...
+		It("the returned resource's template should be the input's unstructured", func() {
+			// This test implicitly tests that the nested hierarchy of fields exists
+			expected, _ = createFederatedResource(scheme, resource)
+
+			mapObj, found, err := getTemplateField(expected)
+			targetResource := &unstructured.Unstructured{}
+
+			convErr := scheme.Convert(resource, targetResource, nil)
+			Expect(convErr).ToNot(HaveOccurred())
+			removeUnwantedFields(targetResource)
```
tpantelis: One thing I try to avoid in UTs is relying on internal aspects of the production code for verification. Here you're using `removeUnwantedFields` to build the expected data which is an internal helper function also used by the code you're testing. So naturally the expected and actual will tend to match even though it might not be correct. So instead, I try to think about what the actual result should be from a blackbox perspective and setup assertions based on that. In this case, what we really expect is for none of the **metadata** fields in the original resource to be present in the federated resource template's **metadata** and thus we can simply verify the associated map is empty:

`Expect(mapObj["metadata"]).To(BeEmpty())`

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,389 @@
...
+	})
+
+	When("the resource type scheme has been added to the type registry", func() {
+		BeforeEach(func() {
+			err := corev1.AddToScheme(scheme)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("should not modify the input resource but copy it", func() {
+			expected, _ = createFederatedResource(scheme, resource)
```
tpantelis: bit of a nit but this should really be named `actual` as `resource` is really the expected data. Same with the other tests below.

pkg/federate/kubefed/resource_test.go
tpantelis: Also for clarity, `mapObj` could be renamed to `actualTemplate`.

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,389 @@
...
+				Expect(err).ToNot(HaveOccurred())
+				expectObj, _, err := getTemplateField(fedResource)
+				Expect(err).ToNot(HaveOccurred())
+
+				Expect(fedPodObj["spec"]).ToNot(BeEquivalentTo(expectObj["spec"]))
+
+				mapObj, _, _ := unstructured.NestedSlice(
+					fedPod.Object,
+					ctlutil.SpecField, ctlutil.TemplateField, "spec", "containers",
+				)
```
tpantelis: For clarity, to follow actual vs expected semantics, some suggested var name changes:

fedPod -> actualResource
fedPodObj -> actualTemplate
mapObj -> actualContainers
image -> actualImage
fedResource -> expResource
expectObj -> expTemplate

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,389 @@
...
+		})
+
+		When("the resource is not in the Kube API", func() {
+			It("should create the resource", func() {
+				// Sanity check
+				_, err := getResourceFromAPI(f.kubeFedClient, fedResource)
+				Expect(err).To(HaveOccurred())
+
+				_ = f.Distribute(resource, clusterNames...)
+				fedPod, err := getResourceFromAPI(f.kubeFedClient, fedResource)
```
tpantelis: Suggested name change: fedPod -> actualResource

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,389 @@
...
+				// This test implicitly tests that the nested hierarchy of fields exists
+				_ = f.Distribute(resource, clusterNames...)
+				fedPod, err := getResourceFromAPI(f.kubeFedClient, fedResource)
+				Expect(err).ToNot(HaveOccurred())
+
+				mapObj, found, err := getClustersField(fedPod)
+				fedClusterNames := unpackClusterNames(mapObj)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(found).To(BeTrue(), "Hierarchy of fields missing!")
+				Expect(fedClusterNames).To(BeEquivalentTo(clusterNames))
```
tpantelis: Suggested name changes: 

fedPod -> actualResource
mapObj -> actualClusters
fedClusterNames -> actualClusterNames
clusterNames -> expClusterNames

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,389 @@
...
+			It("the returned resource's placement should not contain an empty matcher", func() {
+				// This test implicitly tests that the nested hierarchy of fields exists
+				_ = f.Distribute(resource, clusterNames...)
+				fedPod, err := getResourceFromAPI(f.kubeFedClient, fedResource)
+				Expect(err).ToNot(HaveOccurred())
+
+				mapObj, found, err := getMatchLabelsField(fedPod)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(found).ToNot(BeTrue(), "Hierarchy of fields is present!")
+				Expect(mapObj).To(BeEmpty())
```
tpantelis: Similar to above wrt suggested var name changes...

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,389 @@
...
+			It("the returned resource's placement should be an empty matcher", func() {
+				// This test implicitly tests that the nested hierarchy of fields exists
+				_ = f.Distribute(resource)
+				fedPod, err := getResourceFromAPI(f.kubeFedClient, fedResource)
+				Expect(err).ToNot(HaveOccurred())
+
+				mapObj, found, err := getMatchLabelsField(fedPod)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(found).To(BeTrue(), "Hierarchy of fields missing!")
+				Expect(mapObj).To(BeEmpty())
```
tpantelis: Similar to above wrt suggested var name changes...

pkg/federate/kubefed/resource_test.go
mangelajo: Agreed, I was thinking of the same think while writing my other tests this week.

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,389 @@
...
+
+var _ = Describe("Federator", func() {
+	Describe("function Distribute", testDistribute)
+	Describe("helper function createFederatedResource", testCreateFederatedResource)
+})
+
+func testCreateFederatedResource() {
+	var (
+		resource *corev1.Pod
+		expected *unstructured.Unstructured
```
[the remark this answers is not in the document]
mpeterson: yup, leftovers from the refactor.

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,389 @@
...
+
+func (fake *fakeClientWithUpdateError) Create(ctx context.Context, obj runtime.Object) error {
+	return nil
+}
+
+func (fake *fakeClientWithUpdateError) Delete(ctx context.Context, obj runtime.Object, opts ...client.DeleteOptionFunc) error {
+	return nil
+}
+
+func (fake *fakeClientWithUpdateError) Status() client.StatusWriter { return nil }
```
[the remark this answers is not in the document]
mpeterson: right, I still haven't incorporated this about the language.

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,389 @@
...
+				Expect(expected).To(BeNil())
+			})
+		})
+	})
+
+	When("the resource type scheme has been added to the type registry", func() {
+		BeforeEach(func() {
+			err := corev1.AddToScheme(scheme)
+			Expect(err).ToNot(HaveOccurred())
+		})
```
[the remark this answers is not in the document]
mpeterson: >   using scheme.Scheme directly in the places below (and import "k8s.io/client-go/kubernetes/scheme").

That would be really really bad, `scheme.Scheme` is global state and once modified by the `AddToScheme` there is no cleaning up. As a result you affect all tests that come after.

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,389 @@
...
+	BeforeEach(func() {
+		resource = newPod("test-pod", "nginx")
+		scheme = runtime.NewScheme()
+	})
+
+	AfterEach(func() {
+		resource = nil
+		expected = nil
+		scheme = nil
+	})
```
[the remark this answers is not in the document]
mpeterson: I know, but for correctness and future proof I think this is best.

pkg/federate/kubefed/resource_test.go
mpeterson: That would be technically incorrect as it's not only the template field but the whole hierarchy until the template field.

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,389 @@
...
+			Expect(expected.GetNamespace()).
+				To(Equal(resource.GetNamespace()))
+		})
+	})
+}
+
+func testDistribute() {
+	var (
+		resource      *corev1.Pod
+		fedResource   *unstructured.Unstructured
```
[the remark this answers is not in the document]
mpeterson: yes, but I find it technically incorrect to have the var definitions in a nested hierarchy. One example of why this is bad is if in the future someone wants to add `fedResource` for another `When` then they will go and define it there (without refactoring to the most common denominator place). Now you have two places where it's defined and is not clear. Moreover, the variable might be of different types, making things even MORE confusing. 

Also, any savvy programmer will go and look for the variables at the top of the block, as that is good programming pattern.

pkg/federate/kubefed/resource.go
mpeterson: Yes. I want the consumer to have access to it, in case they want to use it.

pkg/federate/kubefed/resource_test.go
mpeterson: same as before, it checks hierarchy and not only the field

pkg/federate/kubefed/resource_test.go
mpeterson: fair enough

pkg/federate/kubefed/resource_test.go
mpeterson: > Can we change this to _"The matchLabels field is missing!"_? Also at the other call sites.

Same as other sites. 

> We could go further and have `getMatchLabelsField` do the checking that's on lines 290-291 and have it return only a valid `mapObj` - that would simplify the call sites. Same with `getClustersField` and `getTemplateField`.

There are variations of the assertions, it's not always the same thing that is expected.

pkg/federate/kubefed/resource_test.go
mpeterson: Sure, just be careful and take in mind the future-proof argument I gave you before and also keep things clear.

pkg/federate/kubefed/resource_test.go
mpeterson: All the ignoring were to keep the tests simpler and to actually assert on the important things which is what the expected outcome is. Those assertions in most cases fail anyways if the ignored bits are ignored. So I see no value on extra cumbersomeness. But adding it.

pkg/federate/kubefed/resource_test.go
mpeterson: You are 100% right. Just be mindful that this and some of your other comments are a bit condescending, Tom.

pkg/federate/kubefed/resource_test.go
mpeterson: You are welcome to change them in future patches.

pkg/federate/kubefed/resource_test.go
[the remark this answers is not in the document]
tpantelis: You don't need to do `AddToScheme` at all on `scheme.Scheme` as it's already built with it.  In this manner you don't have to manipulate any `Scheme` at all.

pkg/federate/kubefed/resource_test.go
tpantelis: Not sure I follow -  you're specifically getting the nested "templates" field here right?

mpeterson:
@mkolesnik @tpantelis I enabled maintainers to add commits to the PR. So, if you have anything else outstanding to change you can add commits before squashing and merging.

How to do it? https://help.github.com/en/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork

tpantelis:
I did my final review and approved. Folks, please do your reviews as soon as you can - I'd like to merge this on Mon if possible.

pkg/federate/kubefed/resource_test.go
```diff
@@ -0,0 +1,372 @@
...
+	ctlutil "sigs.k8s.io/kubefed/pkg/controller/util"
+	kubefedopt "sigs.k8s.io/kubefed/pkg/kubefedctl/options"
+)
+
+type fakeClientWithUpdateError struct {
+	client.Client
+	msg string
+}
+
+var _ client.Client = &fakeClientWithUpdateError{}
```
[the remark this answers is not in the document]
tpantelis: This is done in general to make the compiler check that a struct implements an interface.  It isn't really needed in this case since `fakeClientWithUpdateError` is used as a `client.Client` in code so the compiler will check anyway.

mangelajo:
Feel free to merge, I don't have bandwidth to do another pass.

PR #8, opened 2019-08-07
Use k8s workqueue instead of chan for clusterWatcher

tpantelis:

### DIFF
--- a/pkg/federate/kubefed/cluster.go
+++ b/pkg/federate/kubefed/cluster.go
@@ -14,6 +14,7 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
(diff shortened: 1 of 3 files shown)


pkg/federate/kubefed/cluster.go
```diff
@@ -14,26 +14,48 @@ import (
...
-	// on ClusterEventHandler registration) but not too large to risk running out of memory due to a slow handler.
-	defaultEventQueueBufferSize = 5000
-	defaultEnqueueTimerInterval = time.Second * 10
-)
+type runnable interface {
+	run()
+}
+
+type clusterEvent struct {
+	handler    federate.ClusterEventHandler
```
mkolesnik: Why do you need to keep this as state?

You can instead just pass it as a parameter in the run method since you're running it on the clusterWatcher itself so you already have the event handler at hand.

pkg/federate/kubefed/cluster.go
```diff
@@ -246,62 +261,50 @@ func newKubeFedClusterListerWatcher(kubeFedConfig *rest.Config) (cache.ListerWat
 }
 
 func (c *clusterWatcher) onAdd(clusterID string, kubeConfig *rest.Config) {
```
mkolesnik: Since you already have a struct for the event it makes it much simpler and you can just have one method enqueue which would receive the correct "event struct" and enqueue it.

E.G.
func (c *clusterWatcher) enqueue(event interface{}) {
    c.eventQueue.Add(event)
}

Then for calling when cluster added:
clusterWatcher.enqueue(&clusterAdded{clusterID, kubeConfig})

pkg/federate/kubefed/cluster.go
```diff
@@ -14,26 +14,48 @@ import (
...
 )
 
-const (
-	// The size of the event queue buffer for each ClusterEventHandler. This is largely a heuristic - we want a
-	// value reasonably large enough to avoid blocking the sender on bursts (eg when notifying of initial clusters
-	// on ClusterEventHandler registration) but not too large to risk running out of memory due to a slow handler.
-	defaultEventQueueBufferSize = 5000
-	defaultEnqueueTimerInterval = time.Second * 10
-)
+type runnable interface {
```
mkolesnik: I liked this approach to make the event self runnable, really cool.

pkg/federate/kubefed/cluster.go
```diff
@@ -14,26 +14,48 @@ import (
...
+	run()
+}
+
+type clusterEvent struct {
```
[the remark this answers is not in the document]
tpantelis: It provides a little bit of reuse - I debated its use myself - I can remove it - it doesn't matter to me either way.

pkg/federate/kubefed/cluster.go
tpantelis: sure - either way

pkg/federate/kubefed/cluster.go
```diff
@@ -14,26 +14,48 @@ import (
...
+	clusterEvent
+}
 
 func (f *federator) WatchClusters(handler federate.ClusterEventHandler) error {
 	f.startInformerOnce.Do(func() {
 		go func() {
 			f.kubeFedClusterInformer.Run(f.stopChan)
 			klog.Info("KubeFedCluster informer stopped")
+
+			f.Lock()
```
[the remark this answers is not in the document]
tpantelis: sure

pkg/federate/kubefed/cluster.go
tpantelis: yeah - either way. It's a bit of a remnant of the prior implementation but I think it reads a bit better at the call sites.

tpantelis: Actually thinking about this more, it would be more efficient for the caller to create a single `clusterAdded` instance and pass to each `clusterWatcher` as you suggest instead of each `clusterWatcher` creating it's own instance. Also I'm going to rename `runnable` to 'clusterEvent` for clarity.

pkg/federate/kubefed/cluster.go
```diff
@@ -245,63 +260,31 @@ func newKubeFedClusterListerWatcher(kubeFedConfig *rest.Config) (cache.ListerWat
...
-		case event := <-c.eventQueue:
-			event()
-		case <-c.stopChan:
+		event, shutdown := c.eventQueue.Get()
+		if shutdown {
 			klog.V(2).Infof("ClusterEventHandler \"%T\" eventLoop stopped", c.handler)
 			return
 		}
+
+		event.(clusterEvent).run(c.handler)
```
mkolesnik: No action required: This interface (and generally the handler interface) won't allow for any error. IDK if it's what we'll really want in the future, but we can add it at a later point since it's only tangential to this patch..
Just popped into my mind so I'm writing it here to raise this design decision.

tpantelis: yeah that thought crossed my mind as well.

PR #12, opened 2019-08-14
Admiral CI proposal

dimaunx:


.travis.yml
```diff
@@ -0,0 +1,17 @@
...
+
+services:
+  - docker
+before_script:
+  - CHANGED_FILES_PR=$(git diff --name-only HEAD $(git merge-base HEAD $TRAVIS_BRANCH))
+script:
+  - make ci e2e status=keep
+after_success:
+  - if [[ "${CHANGED_FILES_PR[@]}" =~ "scripts/kind-e2e/e2e.sh" ]]; then
+      echo "scripts/kind-e2e/e2e.sh was modified, testing recurring run on already deployed infrastructure.";
```
mpeterson: what is this supposed to mean?

Dockerfile.dapper
```diff
@@ -0,0 +1,61 @@
+FROM ubuntu:16.04
+
+ARG DAPPER_HOST_ARCH
+ENV HOST_ARCH=${DAPPER_HOST_ARCH} ARCH=${DAPPER_HOST_ARCH} DAPPER_ENV=REPO DAPPER_ENV=TAG \
+    DAPPER_SOURCE=/go/src/github.com/submariner-io/lighthouse DAPPER_DOCKER_SOCKET=true \
```
mpeterson: lighthouse?

Dockerfile.dapper
```diff
@@ -0,0 +1,61 @@
...
+# helm            | e2e tests
+# kubefedctl      | e2e tests
+# kind            | e2e tests
+# ginkgo          | tests
+# goimports       | code formatting
+
+RUN apt-get -q update && \
+    apt-get install -y gcc git curl docker.io make
+
+ENV GO_VERSION=1.12.6
```
mpeterson: isn't it possible to always point to the latest release and to have another file to override versions if needed?

(same comment for all other _VERSION)

scripts/kind-e2e/README.md
```diff
@@ -0,0 +1,111 @@
...
+E2E environment consists of:
+- 3 k8s clusters deployed with [kind].
+  - Cluster1: One master node for broker and elasticsearch if required.
+  - Cluster{2..3}: One master node and two workers for gateways nodes.
+  The configuration can be viewed or changed at **scripts/kind-e2e/cluster{1..3}-config.yaml**.
+- Submariner installed and configured on top of the clusters.
+
+[kind] is a tool for running local Kubernetes clusters using Docker container “nodes”.
+
+## Prerequisites
```
mpeterson: kubefed also, no?

scripts/kind-e2e/README.md
```diff
@@ -0,0 +1,111 @@
...
+make ci e2e
+```
+
+To test specific k8s version, additional **version** parameter can be passed to **make e2e** command.
+
+```bash
+make e2e version=1.13.7
+```
+
+Full list of supported k8s versions can found on [kind release page] page. We are using kind vesion 0.4.0.
```
mpeterson: here and in other places: I don't like hard coding of information such as version numbers. These things will get out-dated in no time. No one will pay attention to update them.

scripts/kind-e2e/README.md
```diff
@@ -0,0 +1,111 @@
+# E2E environment
+
+The e2e environment can be used for local testing, development, and CI purposes.
+
+E2E environment consists of:
+- 3 k8s clusters deployed with [kind].
```
mpeterson: it mentions three clusters but then there are only two configs and the `for` that sets the envs also only brings up two

scripts/kind-e2e/e2e.sh
```diff
@@ -0,0 +1,177 @@
...
+
+### Functions ###
+
+function kind_clusters() {
+    trap_commands
+    status=$1
+    version=$2
+    pids=(-1 -1)
+    logs=()
+    for i in 1 2; do
```
mpeterson: re: only bringing up two

.travis.yml
dimaunx: Testing on recurring runs if we use make e2e status=keep

scripts/kind-e2e/README.md
dimaunx: You are currently , currently we are working on a solution to that.

scripts/kind-e2e/README.md
dimaunx: Added kubefedctl to the readme.

Dockerfile.dapper
dimaunx: Fixed the path to admiral.

scripts/kind-e2e/README.md
dimaunx: Readme fixed.

PR #13, opened 2019-08-15
Add generated Federator mock

tpantelis:
For use with gomock. The file was generated using 'go generate'
via the special commannd directive comment in the code.


mpeterson:
As we talked on the videocall, once the upcoming PR is published for review, this patch will be justified for merging. 

Another option is to just include it on the other PR and we'll merge without squashing. What do you think?

PR #17, opened 2019-08-21
Fix update issue in kubefed Distribute

tpantelis:
For update, a valid ResourceVersion must be specified. The API server/etcd
uses optimistic locking via CAS with the ResourceVersion, ie the input
ResourceVersion must match the existing ResourceVersion on update. The pattern
is to get the existing resource, modify it as needed but retain the ResourceVersion,
then issue the update. There's also a utility, retry.RetryOnConflict, to aid with
retries in case of a conflict due to a concurrent modification.

Also added some debug logging and more context to the error messages to aid
in troubleshooting.

### DIFF
--- a/pkg/federate/kubefed/resource.go
+++ b/pkg/federate/kubefed/resource.go
@@ -3,11 +3,15 @@ package kubefed
 import (
 	"context"
 	"fmt"
+	"reflect"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/util/retry"
+	"k8s.io/klog"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	ctlutil "sigs.k8s.io/kubefed/pkg/controller/util"
 	kubefedopt "sigs.k8s.io/kubefed/pkg/kubefedctl/options"
 )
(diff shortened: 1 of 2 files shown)


pkg/federate/kubefed/resource.go
```diff
@@ -41,16 +46,44 @@ func (f *federator) Distribute(resource runtime.Object, clusterNames ...string)
 	}
 
 	// Update first, as it will probably be the most common operation
-	err = f.kubeFedClient.Update(context.TODO(), fedResource)
+	err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
mpeterson: I think RetryOnConflict should be a consumer decision. Also in our use case there is not going to be need for this and thus its usage is not warranted.

pkg/federate/kubefed/resource.go
```diff
@@ -41,16 +46,44 @@ func (f *federator) Distribute(resource runtime.Object, clusterNames ...string)
...
 	}
 
 	if !errors.IsNotFound(err) {
-		return err
+		return fmt.Errorf("error updating federated resoource %#v: %v", fedResource, err)
+	}
+
+	klog.V(2).Infof("Creating federated resource: %#v", fedResource)
+
+	if err = f.kubeFedClient.Create(context.TODO(), fedResource); err != nil {
```
mpeterson: Since now you are "getting" before issuing an update you know beforehand if there is a need for creation or not. Thus this flow is not needed anymore. You should refactor for readability.

pkg/federate/kubefed/resource.go
```diff
@@ -41,16 +46,44 @@ func (f *federator) Distribute(resource runtime.Object, clusterNames ...string)
...
+		return fmt.Errorf("error creating federated resoource %#v: %v", fedResource, err)
```
mpeterson: too much logging IMHO

pkg/federate/kubefed/resource_test.go
```diff
@@ -223,12 +223,12 @@ func testDistribute() {
 
 		When("on update the Kube API returns an error that is not a NotFound", func() {
 			BeforeEach(func() {
-				clientPatcher = &fakeClientWithUpdateError{fake.NewFakeClient(), "test error"}
+				clientPatcher = &fakeClientWithUpdateError{fake.NewFakeClientWithScheme(scheme, fedResource), "fake error"}
```
mpeterson: Why is this change needed?

pkg/federate/kubefed/resource.go
tpantelis: I think it's safest to do it this way. It certainly doesn't hurt.

pkg/federate/kubefed/resource.go
tpantelis: Not sure I follow...

pkg/federate/kubefed/resource.go
tpantelis: This helps to distinguish between create and update when an error occurs. When I was troubleshooting issues, this was helpful. Any additional context  is always helpful.

pkg/federate/kubefed/resource_test.go
tpantelis: Because in order to test update there needs to be an existing resource.

pkg/federate/kubefed/resource.go
```diff
@@ -41,16 +46,44 @@ func (f *federator) Distribute(resource runtime.Object, clusterNames ...string)
...
+		return f.kubeFedClient.Update(context.TODO(), existing)
+	})
+
 	if err == nil {
 		return nil
 	}
 
 	if !errors.IsNotFound(err) {
-		return err
+		return fmt.Errorf("error updating federated resoource %#v: %v", fedResource, err)
```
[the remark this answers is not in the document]
tpantelis: This helps to distinguish between create and update when an error occurs. When I was troubleshooting issues, this was helpful. Any additional context is always helpful.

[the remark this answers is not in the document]
mpeterson: I will summarize our discussion and decided outcome:

1) We need to weight the advantages of logging disregarding the consumer actions and how that affects double logging if the consumer logs.
2) I'm more proponent of logging here but return the error as is. Tom is for adding the context information to the error string.
3) In general most of the useful information is inside the error attributes but not the string it prints. So it is very useful to introspect with `kubernetes/errors` but less so when we think about solving problems in the future.
4) It's a very philosophical question.

Said that, we agreed with this pattern as it seems to be most used in the code that we are producing as a team.

pkg/federate/kubefed/resource.go
mpeterson: just make sure to cover it with tests

tpantelis:
> Perhaps it makes more sense to use [`CreateOrUpdate`](https://github.com/kubernetes-sigs/controller-runtime/blob/d7467fcf9e1c5817231f2768ffd626a244f2221b/pkg/controller/controllerutil/controllerutil.go#L117) and simplify the logic here. What do you think?
> 

It would make sense to use it but the link above is from the v0.2.0-rc.0 version which looks like the release candidate for the next version.  We're using [v0.1.12](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.1.12/pkg/controller/controllerutil/controllerutil.go). The signature for the `MutateFn` changed and the internal implementation was refactored.  Notice that the v0.1.12  version doesn't immediately check the error from the `Get` call on line 138 although it's not really a big deal.  So they've cleaned up the code in the newer version.

Another thing is that it calls the `MutateFn` on both create and update which means we'd have to either refactor our code a bit wrt `createFederatedResource` (and probably refactor UTs as they test/use it) or keep using `createFederatedResource` as is and incur an additional `DeepCopyObject`.  

Given all that, I'd prefer to keep it as is for now and get this PR in.  We can consider using `CreateOrUpdate` later .

PR #20, opened 2019-08-29
Follow-up for comments in PR 18

tpantelis:
- Refactored usage of removeExistingClusterWatcher
- Rename getLabel/setLabel to getClusterIDLabel/setClusterIDLabel
- Moved SetLabels call
- Refactored structure of runQueueWorker


tpantelis:
The orchestrator code is obsolete now.

PR #22, opened 2019-09-11
Addition of initial E2E tests

mpeterson:
There are several commits that are part of this. I recommend reviewing them individually to make it easier. 

When merged they should not be squashed to retain the history and granularity.

### DIFF
--- a/test/e2e/api/api.go
+++ b/test/e2e/api/api.go
@@ -1,0 +1,112 @@
+package api
+
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/submariner-io/admiral/pkg/federate"
+	"github.com/submariner-io/admiral/test/e2e/api/kubefed"
+	"github.com/submariner-io/admiral/test/e2e/framework"
+)
+
+var controlPlanes = map[string]framework.InitFunc{
+	kubefed.Name: kubefed.InitTestFramework,
+}
+
+var _ = Describe("E2E tests to validate Admiral", func() {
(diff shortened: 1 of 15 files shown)


mpeterson:
This is still a WIP but reviews are welcome and needed!

test/e2e/api/api.go
```diff
@@ -0,0 +1,112 @@
...
+	AfterEach(func() {
+		federator = nil
+		testPod = nil
+		contexts = nil
+	})
+
+	testResourceAPIOperations := func(clusters ...string) {
+		By("Distributing")
+		err := federator.Distribute(testPod, clusters...)
+		Expect(err).ToNot(HaveOccurred())
```
mkolesnik: nit: You can expect it to succeed

test/e2e/api/api.go
```diff
@@ -0,0 +1,112 @@
...
+
+		By("[slow] Checking correct distribution")
+		presence := framework.BuildPresenceMap(contexts, clusters...)
+		framework.AssertPodInClusters(f, testPod, presence)
+
+		By("Deleting")
+		err = federator.Delete(testPod)
+		Expect(err).ToNot(HaveOccurred())
```
mkolesnik: nit: You can expect it to succeed

test/e2e/api/api.go
```diff
@@ -0,0 +1,112 @@
...
+	Context("when we want to distribute to all clusters", func() {
+		testResourceAPI()
+	})
+
+	Context("when we want to distribute to cluster2 only", func() {
+		testResourceAPI("cluster2")
+	})
+}
+
+func newPod(name, namespace string) *v1.Pod {
```
mkolesnik: We're also using newPod in resource_test, perhaps we can extract this to a shared test util class or something

mkolesnik:
This is irrelevant as most or the code has changed substantially

PR #24, opened 2019-10-11
Fix the goStrongswanVici reference

skitt:
Go 1.13 requires longer SHA prefixes, and is stricter with
timestamps. This patch fixes the goStrongswanVici reference
accordingly.

Closes: #23


go.mod
```diff
@@ -24,4 +24,4 @@ require (
 	sigs.k8s.io/testing_frameworks v0.1.1 // indirect
 )
 
-replace github.com/bronze1man/goStrongswanVici => github.com/mangelajo/goStrongswanVici v0.0.0-20190223031456-9a5ae4453bd
+replace github.com/bronze1man/goStrongswanVici => github.com/mangelajo/goStrongswanVici v0.0.0-20190701121157-9a5ae4453bda
```
mangelajo: The replace can be removed now, the patches were properly contributed back to goStrongswanVici :)

skitt: That requires updating the Submariner dependency, is that OK?

tpantelis:
Addressed with https://github.com/submariner-io/admiral/pull/32.

PR #25, opened 2019-10-30
Use the base Dapper container

skitt:


tpantelis:
Addressed by https://github.com/submariner-io/admiral/pull/32.

PR #34, opened 2020-03-25
Refactor Federator API

tpantelis:
- Separated the `WatchClusters` method into a `ClusterInformer` interface
  as the broker component won't provide an implementation for this.
- Removed the `clusterNames` parameter from the `Distribute` method as
  the broker implementation won't have support for this. This wasn't
  used anyway.
- Removed the generated `MockFederator` used with **GoMock**. This really didn't
  provide much value and is not easy to use or maintain. We can provide a
  simple hand-written mock later.


pkg/federate/cluster_informer.go
```diff
@@ -0,0 +1,25 @@
+package federate
+
+import (
+	"k8s.io/client-go/rest"
+)
+
+// ClusterEventHandler handles federated cluster lifecycle event notifications.
+type ClusterEventHandler interface {
+	// OnAdd is called when a cluster is added. The given 'kubeConfig' can be used to access
```
vthapar: Hope we're not going for a design where broker *needs* to access kube API endpoints of joined clusters. This requirement of Kubefed is causing problems with private cluters, so would prefer to avoid this.

pkg/federate/kubefed/cluster.go
```diff
@@ -181,7 +181,7 @@ func (f *federator) OnUpdate(oldObj, newObj interface{}) {
 	}
 }
 
-func (f *federator) buildFederatedClusterConfig(kubeFedCluster *fedv1.KubeFedCluster) (*rest.Config, error) {
+func (f *Federator) buildFederatedClusterConfig(kubeFedCluster *fedv1.KubeFedCluster) (*rest.Config, error) {
```
vthapar: This will change in subsequent patches/PRs?

pkg/federate/kubefed/cluster.go
```diff
@@ -213,7 +213,7 @@ func (f *federator) buildFederatedClusterConfig(kubeFedCluster *fedv1.KubeFedClu
 	return kubeConfig, nil
 }
 
-func (f *federator) initKubeFedClusterInformer(listerWatcher cache.ListerWatcher) {
+func (f *Federator) initKubeFedClusterInformer(listerWatcher cache.ListerWatcher) {
```
vthapar: same as above.

pkg/federate/kubefed/cluster_test.go
```diff
@@ -222,8 +222,8 @@ func newKubeFedCluster(name string) *fedv1.KubeFedCluster {
 	}
 }
 
-func newFederatorWithWatcher(watcher watch.Interface, stopChan <-chan struct{}, initialKubeFedClusters ...fedv1.KubeFedCluster) *federator {
-	federator := &federator{
+func newFederatorWithWatcher(watcher watch.Interface, stopChan <-chan struct{}, initialKubeFedClusters ...fedv1.KubeFedCluster) *Federator {
```
vthapar: same as above

pkg/federate/federator.go
```diff
@@ -4,39 +4,17 @@ package federate
...
 	// failed.
 	//
-	// If clusterNames is not passed, the resource will be distributed to all clusters
-	//
 	// If the resource was previously distributed and the given resource differs, each previous cluster will receive the
-	// updated resource. If previously distributed and the given list of clusters differs, the resource will be
-	// distributed to any new clusters in the updated list and deleted from previous clusters not in the updated list.
-	Distribute(resource runtime.Object, clusterNames ...string) error
+	// updated resource.
+	Distribute(resource runtime.Object) error
```
mkolesnik: Theoretically we could limit the spread of information only to specific clusters even with a broker model, but I agree that it's better to remove if we don't use this anyway.
Only thing I might've changed is the method name to `DistributeOnAllClusters` or something of this sort, since go doesn't have method overloading.

pkg/federate/cluster_informer.go
tpantelis: The broker implementation won't use the interfaces in this file at all as it has no need to be informed of cluster lifecycle. And neither will lighthouse need this interface. We could just remove this file altogether.

pkg/federate/federator.go
tpantelis: yeah I'd rather not include a feature in the API that won't be implemented or used going forward (although it wasn't used before anyway). We can always add it back later if needed.

The fact that it distributes to all clusters is mentioned in the doc comments, thus I don't think the name needs to specifically indicate that for brevity.

pkg/federate/kubefed/cluster.go
tpantelis: The kubefed implementation won't change - in fact we'll likely end up removing at some point.

PR #43, opened 2020-03-30
Add signals code from submariner

tpantelis:
This code is reusable amongst the submariner-io projects and will
eliminate the submariner dependency in lighthouse.


tpantelis:
I just discovered that the same signals code exists in controller-runtime so no need to maintain a copy.

PR #45, opened 2020-03-31
Move workqueue from submariner for reuse

tpantelis:
The idea here is to put reusable components into admiral to be used by submariner and lighthouse (and possibly others in the future). This avoids lighthouse having to depend on submariner and its release cadence. 

Another reusable component will be the generic datastoresyncer which will live in admiral.  This component will also want to use the workqueue wrapper - if we left it in submariner then we'd have a cross-dependency between the projects.


skitt:
What’s the dependency tree supposed to look like ultimately? Couldn’t admiral import these from submariner? (I imagine there’s a good reason not to, but it would be worth explaining in the commit message.)

tpantelis:
> What’s the dependency tree supposed to look like ultimately? Couldn’t admiral import these from submariner? (I imagine there’s a good reason not to, but it would be worth explaining in the commit message.)

I updated the summary.

PR #46, opened 2020-04-06
Add broker Federator implementation

tpantelis:

### DIFF
--- a/pkg/syncer/broker/federator.go
+++ b/pkg/syncer/broker/federator.go
@@ -7,6 +7,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/federate"
 	"github.com/submariner-io/admiral/pkg/log"
+	"github.com/submariner-io/admiral/pkg/util"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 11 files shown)


pkg/syncer/broker/federator.go
```diff
@@ -0,0 +1,166 @@
...
+		if apierrors.IsNotFound(err) {
+			removeFieldsForCreate(toDistribute)
+			toDistribute.SetNamespace(f.brokerNamespace)
+
+			klog.V(log.DEBUG).Infof("Creating resource: %#v", toDistribute)
+
+			_, err := resourceClient.Create(toDistribute, metav1.CreateOptions{})
+			if apierrors.IsAlreadyExists(err) {
+				klog.V(log.DEBUG).Infof("Resource %q already exists", toDistribute.GetName())
+				return nil
```
skitt, line 89: Don’t we want to update it here? (I mean, fall through to the update code.)

tpantelis: I thought of that but this means it was created out-of-band so unsure whether we should automatically replace it or if the caller would want to merge in this case. Perhaps we should just return the error and let the caller decide how to handle it. If they want to replace then they can retry the call. In our uses cases this `shouldn't` happen.

tpantelis: Thinking about this again, I think we should just update here. The caller expects its version of the resource to be applied if it succeeds and I think returning `AlreadyExists` would be confusing. This is just a timing edge case that is really equivalent to the preceding `Get` returning success a split second earlier if the resource was created out-of-band.

tpantelis: So I'll return a conflict error to cause a retry.

tpantelis: Done

PR #47, opened 2020-04-10
Add generic resource syncer

tpantelis:
This is a series of commits to add a component that watches resources of a specific type and syncs them to a Federator instance. It uses the `federate.ClusterIDLabelKey` to avoid sync loops based on the given local cluster ID and sync direction, either local -> remote or remote -> local.


pkg/syncer/broker/broker_suite_test.go
```diff
@@ -5,8 +5,13 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"k8s.io/klog"
 )
 
+var _ = Describe("", func() {
+	klog.InitFlags(nil)
```
mangelajo, line 12: Why this here with an empty "Describe" ?, wouldn't make more sense to just add inside "TestBroker" ?

pkg/syncer/syncer_suite_test.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"k8s.io/klog"
+)
+
+var _ = Describe("", func() {
+	klog.InitFlags(nil)
```
mangelajo, line 12: same, can we move this down to "TestSyncer" instead?

pkg/syncer/broker/broker_suite_test.go
tpantelis: This placement of this is tricky - it doesn't work if it's in the main Test* function. It also doesn't work if placed inside a `BeforeSuite`. It seems to only works when inside a test spec, ie `Describe`, `Context` etc. And it can only be called once. So if there's multiple top-level `Describe`'s in multiple test files (there isn't now but there will be) then it will fail. So I place it in a single "dummy" top-level `Describe`, with an empty string so nothing is printed.

mangelajo: Ooops, 

May be we can add a comment explaining? :) ':D

tpantelis: I moved `InitFlags` to an `init` function instead with https://github.com/submariner-io/admiral/pull/47/commits/3f781aa06003955e66530136e05c657b50f8c0d3.

PR #53, opened 2020-05-07
Add validations as GH actions

### LINKED ISSUE #52: Implement github actions / admiral

mkolesnik:
Closes #52


tpantelis:
I've since added e2e tests so we should add a an action for that as well.

tpantelis:
Admiral has e2e now so we should add an action for that but that can be done in a separate PR.

PR #68, opened 2020-06-17
Check whether the CA trust chain is needed

skitt:
The trust anchor stored in the broker information reflects the trust
chain of the token, not necessarily the trust chain of the API
endpoint; but we're using it for the latter in all cases.

This changes NewSyncer to check whether the clientset can be used,
by retrieving the first resource that's supposed to be synchronised.
This allows the configuration to be verified, and switched to use the
stored trust anchor if necessary.

### DIFF
--- a/pkg/syncer/broker/config.go
+++ b/pkg/syncer/broker/config.go
@@ -16,7 +16,7 @@ type brokerSpecification struct {
 	Ca              string
 }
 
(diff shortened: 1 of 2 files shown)


skitt:
The verification for this will be in the corresponding Lighthouse patch to bump the Admiral dependency (currently running e2e locally).

skitt:
https://github.com/submariner-io/lighthouse/pull/168

PR #84, opened 2020-07-09
Add callback on successful sync

### LINKED ISSUE #66: Add pre/post callbacks to sync API
Lighthouse uses sync broker API to sync resources across cluster when a ServiceExport is created on source. To correctly update the status we need some notification when a resource is synced to the broker, at very least.

Additional callbacks for each cluster we sync to will be welcome for more information in status.

tpantelis:
Closes #66

### DIFF
--- a/pkg/syncer/resource_syncer.go
+++ b/pkg/syncer/resource_syncer.go
@@ -56,6 +56,9 @@ func (o Operation) String() string {
 // to be retried later.
 type TransformFunc func(from runtime.Object, op Operation) (runtime.Object, bool)
 
+// OnSuccessfulSyncFunc is invoked after a successful sync operation.
+type OnSuccessfulSyncFunc func(synced runtime.Object, op Operation)
+
 type ResourceSyncerConfig struct {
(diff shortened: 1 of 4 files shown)


pkg/syncer/resource_syncer.go
```diff
@@ -175,7 +192,7 @@ func (r *resourceSyncer) processNextWorkItem(key, name, ns string) (bool, error)
 	klog.V(log.DEBUG).Infof("Syncer %q retrieved %sd resource: %#v", r.config.Name, op, resource)
 
 	if r.shouldSync(resource) {
-		resource, requeue := r.transform(resource, op)
+		resource, transformed, requeue := r.transform(resource, op)
```
vthapar, line 195: What is this new value for? Currently resource contains the transformed object. With this change is the expectation for resource to be original and transformed to be new?

tpantelis: yes - there may not be a transformed resource if there's no callback for it, in which case the successful sync callback will use the original resource.

PR #92, opened 2020-07-16
Enable golint and stylecheck linting, fix errors

dfarrell07:
Both linters flag the same dot imports in this repo, so fixing one fixes both.

Builds on https://github.com/submariner-io/admiral/pull/87.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -9,6 +9,8 @@ linters-settings:
       - unlambda
   gocyclo:
     min-complexity: 15
+  golint:
+    min-confidence: 0
   lll:
     line-length: 140
   maligned:
@@ -29,7 +31,7 @@ linters:
     - gocyclo
     - gofmt
     - goimports
-    #- golint
+    - golint
     - gosec
     - gosimple
     - govet
@@ -41,7 +43,7 @@ linters:
     - nakedret
     - staticcheck
     - structcheck
-    #- stylecheck
+    - stylecheck
     #- testpackage
     - typecheck
     - unconvert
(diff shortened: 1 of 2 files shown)


pkg/fake/dynamic_client.go
```diff
@@ -3,7 +3,7 @@ package fake
 import (
 	"time"
 
-	. "github.com/onsi/gomega"
+	g "github.com/onsi/gomega"
```
tpantelis, line 6: In general dot import is discouraged but an exception is made for ginkgo and gomega otherwise it kind of detracts from the fluency of BDD. Is there anyway we can exclude dot import?

dfarrell07: Yeah, I was torn about this one too. I also like the simplicity of the directly imported functions, but as I thought about it more it didn't seem to bad to have to qualify them like normal objects.

There are some other ways of avoiding dot imports described here, but they are a bit more complex and they also kinda-endorse this more verbose method.

https://onsi.github.io/ginkgo/#avoiding-dot-imports

tpantelis: yeah creating top-level `vars` to effect dot import seems a bit overboard to me just to explicitly avoid dot import. I still prefer dot import in these 2 cases - both for the fluency aesthetics and ease of auto-complete in the IDE. In other code I've seen using gingko, it does seem tom be the norm.

dfarrell07: Per discussion on Automation call we're waiting on @skitt and @mangelajo to get back from PTO before making this decision, as we suspect they have thoughts and if so they should definitely be heard first.

skitt: Idiomatic Ginkgo relies on the dot-imports, I think it would be a shame to lose that...

dfarrell07: Okay, so I'll propose an update with an exception to ignore Ginkgo and Gomega dot imports.

PR #95, opened 2020-07-22
Add exported broker syncer constructor for UTs

tpantelis:
Users to be able to construct a broker syncer with fake clients for UTs. There is a non-exported `newSyncer` function used by the local UTs so export it. This enables the local syncer UTs to be moved to the test package.
    
 Also added some more UT utilities.

### DIFF
--- a/pkg/syncer/broker/syncer.go
+++ b/pkg/syncer/broker/syncer.go
@@ -54,6 +54,9 @@ type SyncerConfig struct {
 
 	// ResourceConfigs the configurations for resources to sync
 	ResourceConfigs []ResourceConfig
+
+	// Scheme used to convert resource objects. By default the global k8s Scheme is used.
+	Scheme *runtime.Scheme
 }
 
 type Syncer struct {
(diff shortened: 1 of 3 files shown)


pkg/syncer/broker/syncer.go
```diff
@@ -189,11 +188,6 @@ func (s *Syncer) Start(stopCh <-chan struct{}) error {
...
 }
 
-func (s *Syncer) GetBrokerFederatorFor(resourceType runtime.Object) federate.Federator {
-	f, found := s.federators[reflect.TypeOf(resourceType)]
-	if found {
-		return f
-	}
-
-	return nil
+func (s *Syncer) GetBrokerFederator() federate.Federator {
```
vthapar, line 191: This looks like a change that belongs to #86 I believe merge and rebase will take care of this, right?

tpantelis: I based https://github.com/submariner-io/admiral/pull/95/commits/db1696ad5673e62afb375741ff475c9e5e98d602 on that commit to avoid conflict.

PR #97, opened 2020-07-29
Bump to Kubernetes 1.17.0

skitt:
... to match Shipyard.


skitt:
https://github.com/kubernetes/client-go/compare/kubernetes-1.17.0-alpha.3..kubernetes-1.17.0-beta.0 shows what changed in client-go between the two releases around the breakage, but I can’t figure out why beta 0 breaks Admiral...

Issue #115, opened 2020-08-31
Add method to get cached resource from syncer


tpantelis:
Lighthouse has several syncers and transformers that access resources across syncers from the API server. It would be beneficial to expose a method to retrieve a resource from a syncer's informer cache for efficiency.

PR #116, opened 2020-08-31
Add method to get cached resource from syncer

tpantelis:
See https://github.com/submariner-io/admiral/issues/115

### DIFF
--- a/pkg/syncer/resource_syncer_test.go
+++ b/pkg/syncer/resource_syncer_test.go
@@ -36,6 +36,7 @@ var _ = Describe("Resource Syncer", func() {
 	Describe("With OnSuccessfulSync Function", testOnSuccessfulSyncFunction)
 	Describe("Sync Errors", testSyncErrors)
 	Describe("Update Suppression", testUpdateSuppression)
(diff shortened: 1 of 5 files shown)


pkg/syncer/resource_syncer_test.go
```diff
@@ -533,6 +534,35 @@ func testUpdateSuppression() {
...
 
+func testGetResource() {
+	d := newTestDiver(test.LocalNamespace, "", syncer.LocalToRemote)
+
+	When("the requested resource exists", func() {
+		BeforeEach(func() {
+			d.addInitialResource(d.resource)
+		})
+
+		It("should successfully the resource", func() {
```
skitt, line 545:
```suggestion
		It("should return the resource", func() {
```

skitt (changes requested):

pkg/syncer/resource_syncer_test.go
tpantelis: yup 😄

skitt (approved):

PR #125, opened 2020-09-28
Make metadata fields to preserve configurable in broker Federator

tpantelis:
There are cases where additional fields need to be preserved so make it configurable. An example is `OwnerReferences` when syncing local -> local.

### DIFF
--- a/pkg/syncer/broker/federator.go
+++ b/pkg/syncer/broker/federator.go
@@ -14,22 +14,28 @@ import (
 )
 
 type federator struct {
-	dynClient       dynamic.Interface
-	restMapper      meta.RESTMapper
-	targetNamespace string
-	localClusterID  string
+	dynClient          dynamic.Interface
+	restMapper         meta.RESTMapper
+	targetNamespace    string
+	localClusterID     string
+	keepMetadataFields map[string]bool
 }
 
-var keepMetadataFields = map[string]bool{"name": true, "namespace": true, util.LabelsField: true, "annotations": true}
-
 func NewFederator(dynClient dynamic.Interface, restMapper meta.RESTMapper, targetNamespace,
-	localClusterID string) federate.Federator {
-	return &federator{
-		dynClient:       dynClient,
-		restMapper:      restMapper,
-		targetNamespace: targetNamespace,
-		localClusterID:  localClusterID,
+	localClusterID string, keepMetadataField ...string) federate.Federator {
+	f := &federator{
+		dynClient:          dynClient,
+		restMapper:         restMapper,
+		targetNamespace:    targetNamespace,
+		localClusterID:     localClusterID,
+		keepMetadataFields: map[string]bool{"name": true, "namespace": true, util.LabelsField: true, "annotations": true},
 	}
+
+	for _, field := range keepMetadataField {
+		f.keepMetadataFields[field] = true
+	}
+
+	return f
 }
 
 func (f *federator) Distribute(resource runtime.Object) error {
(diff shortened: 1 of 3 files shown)


skitt (approved):
I’ve checked the behaviour when setting owner or controller references on synchronised objects; the owner check compares GVK and name, not the uids, so this doesn’t prevent updating the owner or controller on the target object.

tpantelis:
> I’ve checked the behaviour when setting owner or controller references on synchronised objects; the owner check compares GVK and name, not the uids, so this doesn’t prevent updating the owner or controller on the target object.

I'm sure I follow - what owner check? The point of this change is to let the user at a higher level control whether or not `OwnerRefs` are transferred.  The specific case is `EndpointSlices` in LH where we use a syncer for local `Endpoints` -> local `EndpointSlice` - in this case we want the `Federator` to preserve the `OwnerRef`. When syncing remotely we don't want to propagate `OwnerRef` so a transform func is used to strip out the `OwnerRef` for local `EndpointSlice` -> broker.

skitt:
> > I’ve checked the behaviour when setting owner or controller references on synchronised objects; the owner check compares GVK and name, not the uids, so this doesn’t prevent updating the owner or controller on the target object.
> 
> I'm sure I follow - what owner check? The point of this change is to let the user at a higher level control whether or not `OwnerRefs` are transferred. The specific case is `EndpointSlices` in LH where we use a syncer for local `Endpoints` -> local `EndpointSlice` - in this case we want the `Federator` to preserve the `OwnerRef`. When syncing remotely we don't want to propagate `OwnerRef` so a transform func is used to strip out the `OwnerRef` for local `EndpointSlice` -> broker.

Yeah, I should have explained in more detail. I was concerned that the default case would cause problems because of `controllerutil`’s checks when setting the owner reference: if an object is synced remotely, and the owner reference isn’t stripped, I wondered whether we could end up with an object whose owner reference can’t be fixed (at least, not through `controllerutil`). I’m assuming that if we’re syncing trees of objects, the owners will be synced too, or at least they’ll have the same GVK and name.

tpantelis:
> 
> I'm sure I follow - what owner check? The point of this change is to let the user at a higher level control whether or not `OwnerRefs` are transferred. The specific case is `EndpointSlices` in LH where we use a syncer for local `Endpoints` -> local `EndpointSlice` - in this case we want the `Federator` to preserve the `OwnerRef`. When syncing remotely we don't want to propagate `OwnerRef` so a transform func is used to strip out the `OwnerRef` for local `EndpointSlice` -> broker.

Thinking some more, perhaps it's better to keep the default behavior to strip out `OwnerRef` - when syncing remotely, I don't think you'd ever want to preserve the `OwnerRef`. The use case described above is a bit special in that I'm customizing a local -> local syncer.  I think it would be better to configure the `Federator`  to keep the `OwnerRef`.

skitt:
Incidentally, this:

> The point of this change is to let the user at a higher level control whether or not `OwnerRefs` are transferred. The specific case is `EndpointSlices` in LH where we use a syncer for local `Endpoints` -> local `EndpointSlice` - in this case we want the `Federator` to preserve the `OwnerRef`. When syncing remotely we don't want to propagate `OwnerRef` so a transform func is used to strip out the `OwnerRef` for local `EndpointSlice` -> broker.

is the sort of information that would be great to preserve in the commit message!

tpantelis:
> Yeah, I should have explained in more detail. I was concerned that the default case would cause problems because of `controllerutil`’s checks when setting the owner reference: if an object is synced remotely, and the owner reference isn’t stripped, I wondered whether we could end up with an object whose owner reference can’t be fixed (at least, not through `controllerutil`). I’m assuming that if we’re syncing trees of objects, the owners will be synced too, or at least they’ll have the same GVK and name.

yeah I think it would be better to keep the previous default behavior per my subsequent comment.

skitt (approved):

PR #148, opened 2020-11-04
Fix bug in the resource watcher

tpantelis:
We need to capture` rc.Handler` in a closure var within the loop otherwise the next iteration overwrites `rc` and the function passed to the `ResourceSyncer` observes the incorrect handler.

Also moved the Name field to the `ResourceConfig`.


mangelajo (approved):
Thank you for handling this so quickly

PR #165, opened 2020-12-17
Populate StringSet on creation

tpantelis:
Allow initial strings on creation and new `AddAll` function for convenience.

### DIFF
--- a/pkg/stringset/stringset.go
+++ b/pkg/stringset/stringset.go
@@ -6,6 +6,7 @@ import (
 
 type Interface interface {
 	Add(s string) bool
+	AddAll(s ...string)
 	Remove(s string) bool
 	RemoveAll()
 	Contains(s string) bool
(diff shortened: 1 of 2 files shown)


dfarrell07 (approved):

pkg/stringset/stringset.go
```diff
@@ -23,16 +24,19 @@ type synchronized struct {
...
-func NewSynchronized() Interface {
-	return &synchronized{setType: newSetType()}
+func NewSynchronized(s ...string) Interface {
+	return &synchronized{setType: newSetType(s...)}
 }
 
-func newSetType() *setType {
-	return &setType{set: make(map[string]bool)}
+func newSetType(s ...string) *setType {
+	set := &setType{set: make(map[string]bool)}
```
skitt, line 36:
```suggestion
	set := &setType{set: make(map[string]bool, len(s))}
```

skitt (changes requested):
Just a minor tweak which will only make a difference for large string sets, but I think it’s worth doing now...

pkg/stringset/stringset.go
tpantelis: good idea!

skitt (approved):

PR #178, opened 2021-02-06
Use shared GHAs from Shipyard

dfarrell07:
Run common Submariner CI logic like E2E tests, upgrade tests, and the
flake finder using Submariner's custom GitHub Actions from Shipyard.

Signed-off-by: Daniel Farrell <dfarrell@redhat.com>

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -14,5 +14,8 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Run E2E deployment and tests
-        run: |
-          make e2e
+        uses: submariner-io/shipyard/gh-actions/e2e@devel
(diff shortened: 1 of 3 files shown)


skitt (approved):

dfarrell07:
@skitt @tpantelis @mangelajo I'm allowed to merge this after only Stephen's review, which I think shouldn't be the case. Maybe someone can take a look at the GH config?

skitt:
> @skitt @tpantelis @mangelajo I'm allowed to merge this after only Stephen's review, which I think shouldn't be the case. Maybe someone can take a look at the GH config?

Fixed, thanks!

tpantelis (approved):

Issue #136, opened 2020-10-21
Admiral/Broker Prometheus metrics


maayanf24:
Add prometheus metrics for broker
See here for details: https://github.com/submariner-io/submariner/issues/678

admiral library statistics: distributing / getting objects

- [ ] broker response times

- [x] received objects -  [#179 ](https://github.com/submariner-io/admiral/pull/179) and [#184 ](https://github.com/submariner-io/admiral/pull/184#event-4358612732)

- [x] sent objects -  [#179 ](https://github.com/submariner-io/admiral/pull/179) and [#184 ](https://github.com/submariner-io/admiral/pull/184#event-4358612732)

PR #179, opened 2021-02-08
Resource syncer metrics

maayanf24:
Related issue #136

Signed-off-by: Maayan Friedman <maafried@redhat.com>


pkg/syncer/resource_syncer.go
```diff
@@ -67,6 +79,12 @@ func (o Operation) String() string {
 	}
 }
 
+const (
+	directionLabel  = "direction"
+	operationLable  = "operation"
```
skitt:
```suggestion
	operationLabel  = "operation"
```

pkg/syncer/resource_syncer.go
```diff
@@ -137,16 +155,20 @@ type ResourceSyncerConfig struct {
...
-	created   sync.Map
-	stopped   chan struct{}
+	workQueue   workqueue.Interface
+	informer    cache.Controller
+	store       cache.Store
+	config      ResourceSyncerConfig
+	deleted     sync.Map
+	created     sync.Map
+	stopped     chan struct{}
+	metricGauge *prometheus.GaugeVec
```
skitt: Instead of storing the type in the member name, and indicating only that it’s a metric, let’s describe what it’s used for:

```suggestion
	syncCounter *prometheus.GaugeVec
```

skitt (changes requested):
@tpantelis what do you think of this? Is it OK to tie Admiral to Prometheus?

skitt (approved):

mangelajo (approved):

tpantelis (approved):

PR #184, opened 2021-02-16
Create a metric gauge in broker.syncer and send it to resource syncer

maayanf24:
If a resource syncer is created from broker.syncer, create one gauge
on the broker and pass it to use in resource syncer

Related issue #136

Signed-off-by: Maayan Friedman <maafried@redhat.com>

### DIFF
--- a/pkg/syncer/resource_syncer.go
+++ b/pkg/syncer/resource_syncer.go
@@ -156,11 +156,13 @@ type ResourceSyncerConfig struct {
 	// ResyncPeriod if non-zero, the period at which resources will be re-synced regardless if anything changed. Default is 0.
 	ResyncPeriod time.Duration
 
-	// MetricOpts used to pass name and help text to resource syncer Gauge
+	// MetricOpts if specified, used to create a gauge to record counter metrics.
+	// Alternatively the gauge can be created directly and passed via the MetricGauge field,
+	// in which case MetricOpts is ignored.
 	MetricOpts *prometheus.GaugeOpts
 
-	// When used from broker.syncer, SyncCounter will be shared by several resourceSyncer
-	SyncCounter *prometheus.GaugeVec
+	// MetricGauge if specified, used to record counter metrics.
+	MetricGauge *prometheus.GaugeVec
 }
 
 type resourceSyncer struct {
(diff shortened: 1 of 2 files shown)


pkg/syncer/resource_syncer.go
```diff
@@ -158,6 +158,9 @@ type ResourceSyncerConfig struct {
 
 	// MetricOpts used to pass name and help text to resource syncer Gauge
 	MetricOpts *prometheus.GaugeOpts
+
+	// When used from broker.syncer, SyncCounter will be shared by several resourceSyncer
+	SyncCounter *prometheus.GaugeVec
```
tpantelis:
```suggestion
	// MetricGauge if specified, used to record counter metrics.
	MetricGauge *prometheus.GaugeVec
```

pkg/syncer/resource_syncer.go
```diff
@@ -158,6 +158,9 @@ type ResourceSyncerConfig struct {
 
 	// MetricOpts used to pass name and help text to resource syncer Gauge
 	MetricOpts *prometheus.GaugeOpts
```
tpantelis:
```suggestion
	// MetricOpts if specified, used to create a gauge to record counter metrics. Alternatively the gauge can be created directly and passed via the MetricGauge field, in which case MetricOpts is ignored. 
	MetricOpts *prometheus.GaugeOpts
```

pkg/syncer/resource_syncer.go
```diff
@@ -199,12 +202,14 @@ func NewResourceSyncer(config *ResourceSyncerConfig) (Interface, error) {
...
-				directionLabel,
-				operationLabel,
-				syncerNameLabel,
+				DirectionLabel,
+				OperationLabel,
+				SyncerNameLabel,
 			},
 		)
 		prometheus.MustRegister(syncer.syncCounter)
+	} else if syncer.config.SyncCounter != nil {
```
tpantelis: I think this should take precedence over `MetricOpts`.

tpantelis (changes requested):

pkg/syncer/resource_syncer.go
skitt: I requested the `SyncCounter` name; `MetricGauge` doesn’t contain any useful information IMO — we know it’s a metric because it’s a Prometheus object, and we know it’s a gauge because it’s a `GaugeVec`, and featuring both of those in the field name ties it to its type. I’d rather the name reflect the use.

tpantelis: ok - makes sense. Perhaps we should rename `MetricOpts` to `SyncCounterOpts`.  The main reason I suggested `MetricGauge` was to align it with `MetricOpts`.

maayanf24: @skitt - Right, I forgot about that..

tpantelis (approved):

skitt (approved):

PR #185, opened 2021-02-23
Add CreateAnew function

tpantelis:
For use cases where an existing instance needs to first be deleted.

Also added unit tests for the new function and the existing `CreateOrUpdate` function.

### DIFF
--- a/pkg/util/create_or_update.go
+++ b/pkg/util/create_or_update.go
@@ -16,14 +16,19 @@ limitations under the License.
 package util
 
 import (
+	"time"
+
 	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/log"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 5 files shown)


pkg/util/create_or_update.go
```diff
@@ -88,3 +101,32 @@ func CreateOrUpdate(client dynamic.ResourceInterface, obj *unstructured.Unstruct
...
+
+// CreateAnew creates a resource, first deleting an existing instance if one exists.
+func CreateAnew(client dynamic.ResourceInterface, obj runtime.Object) error {
+	toCreate := &unstructured.Unstructured{}
+	err := scheme.Scheme.Convert(obj, toCreate, nil)
+	if err != nil {
+		return err
+	}
+
+	return retry.OnError(backOff, apierrors.IsAlreadyExists, func() error {
```
skitt: Is there any point in retrying if `Create()` returns an error other than “already exists”?

skitt: OK, for anyone else following along, @tpantelis pointed out to me that the function is only retried if the error is accepted by the function given as the second argument to `OnError`, _i.e._ `IsAlreadyExists`.

tpantelis: I refactored it to use `wait.ExponentialBackoff` directly as `OnError` doesn't really add value.

skitt (approved):

PR #187, opened 2021-02-24
Add a CreateAnew variant waiting for deletion

skitt:
This is required for callers who want to ensure that deletion
completes fully (within a timeout) before the object is created anew.

### DIFF
--- a/pkg/util/create_or_update.go
+++ b/pkg/util/create_or_update.go
@@ -105,7 +105,11 @@ func CreateOrUpdate(client dynamic.ResourceInterface, obj *unstructured.Unstruct
 }
 
 // CreateAnew creates a resource, first deleting an existing instance if one exists.
-func CreateAnew(client dynamic.ResourceInterface, obj runtime.Object) error {
+// If the delete options specify that deletion should be propagated in the foreground,
+// this will wait for the deletion to be complete before creating the new object:
+// with foreground propagation, Get will continue to return the object being deleted
+// and Create will fail with “already exists” until deletion is complete.
+func CreateAnew(client dynamic.ResourceInterface, obj runtime.Object, deleteOptions *metav1.DeleteOptions) error {
 	toCreate := &unstructured.Unstructured{}
 	err := scheme.Scheme.Convert(obj, toCreate, nil)
 	if err != nil {
@@ -118,7 +122,7 @@ func CreateAnew(client dynamic.ResourceInterface, obj runtime.Object) error {
 			return true, err
 		}
 
-		err = client.Delete(toCreate.GetName(), &metav1.DeleteOptions{})
+		err = client.Delete(toCreate.GetName(), deleteOptions)
 		if apierrors.IsNotFound(err) {
 			err = nil
 		}
--- a/pkg/util/create_or_update_test.go
+++ b/pkg/util/create_or_update_test.go
@@ -27,6 +27,7 @@ import (
 	"github.com/submariner-io/admiral/pkg/util"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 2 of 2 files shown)


pkg/util/create_or_update.go
```diff
@@ -118,11 +124,25 @@ func CreateAnew(client dynamic.ResourceInterface, obj runtime.Object) error {
...
+		// Wait for the deletion to complete
+		if err == nil && timeout.Nanoseconds() != 0 {
+			err = wait.PollImmediate(1*time.Second, timeout, func() (bool, error) {
+				_, err := client.Get(toCreate.GetName(), metav1.GetOptions{})
+				if apierrors.IsNotFound(err) {
+					return true, nil
+				}
+				return false, err
+			})
+		}
```
tpantelis: It doesn't seem like we actually need this part. If `Get` still returns the instance due to deletion not complete, then I'd expect `Create` to return `IsAlreadyExists` in which case this is handled automatically by the `ExponentialBackoff` loop. I'm not sure what happens if another `Delete` is issued when already in progress - I'd expect it to be a noop. If that's an issue then we can issue the `Get` prior to `Create` above.

skitt: Indeed, and that would avoid having to declare a second function. I’ll try it out in the operator...

skitt: This is fantastic — it ends up retrying multiple times and deleting and creating the CR over and over again, which leaves the operator in a rather interesting state!

skitt: But this version doesn’t work either.

tpantelis: How does that happen...what does Create return after the Delete?

skitt: OK, I think I know what’s going on — I need to fix the operator to ignore Submariner CRs which are being deleted. `Get()` returns the Submariner CR which is being deleted, which messes up the reconciliation loop if deletion takes just a little too long. As a result, resources keep being added with the deleted Submariner CR as controller, so the deletion never completes; the loop times out, and the second creation never happens.

tpantelis: ahh so the problem is in the reconciliation  loop and not the `CreateAnew` loop.  I assume we can check the `DeletionTimestamp` and/or `Finalizers`.

pkg/util/create_or_update.go
```diff
@@ -106,6 +106,16 @@ func CreateOrUpdate(client dynamic.ResourceInterface, obj *unstructured.Unstruct
...
+	return createAnewMayWait(client, obj, false)
+}
+
+// CreateAnewWait creates a resource, first deleting an existing instance if one exists,
+// waiting for the deletion to complete.
+func CreateAnewWait(client dynamic.ResourceInterface, obj runtime.Object) error {
+	return createAnewMayWait(client, obj, true)
+}
+
+func createAnewMayWait(client dynamic.ResourceInterface, obj runtime.Object, waitForDeletion bool) error {
```
tpantelis: I think we can just call this `createAnew`.  In fact I suggest to just add the param to the public function  and keep it `CreateAnew` rather than adding a second function.  Also we can make this even more general by making the param `*metav1.DeleteOptions`,  like the k8s `Delete` method - than we wouldn't have to change the API again if the need for other delete options arises.

pkg/util/create_or_update.go
skitt: Yes, see https://github.com/submariner-io/submariner-operator/pull/1115.

pkg/util/create_or_update.go
```diff
@@ -105,7 +105,12 @@ func CreateOrUpdate(client dynamic.ResourceInterface, obj *unstructured.Unstruct
 }
 
 // CreateAnew creates a resource, first deleting an existing instance if one exists.
-func CreateAnew(client dynamic.ResourceInterface, obj runtime.Object) error {
+// If the delete options specify that deletion should be propagated in the foreground,
+// this will wait for the deletion to be complete before creating the new object
+// (with foreground propagation, Get() will continue returning the object being deleted
+// until all its finalizers have completed, so Create() will fail with “already exists”
+// until deletion is complete).
```
tpantelis:
```suggestion
// (with foreground propagation, Get will continue to return the object being deleted and Create will fail with
// “already exists” until deletion is complete).
```

tpantelis (commented):
Just minor changes to the comment.

pkg/util/create_or_update.go
dfarrell07: Also it might be good to break this very long parentheses section out into a new sentence.

tpantelis (approved):

PR #207, opened 2021-04-14
Enable stylecheck, dogsled, prealloc and goconst linters

Jaanki:
Signed-off-by: Janki Chhatbar <jchhatba@redhat.com>

### DIFF
--- a/pkg/syncer/resource_syncer.go
+++ b/pkg/syncer/resource_syncer.go
@@ -290,9 +290,10 @@ func (r *resourceSyncer) GetResource(name, namespace string) (runtime.Object, bo
 }
 
 func (r *resourceSyncer) ListResources() ([]runtime.Object, error) {
-	var retObjects []runtime.Object
+	list := r.store.List()
+	retObjects := make([]runtime.Object, 0, len(list))
 
-	for _, obj := range r.store.List() {
+	for _, obj := range list {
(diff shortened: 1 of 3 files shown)


pkg/syncer/resource_syncer.go
```diff
@@ -290,7 +290,7 @@ func (r *resourceSyncer) GetResource(name, namespace string) (runtime.Object, bo
 }
 
 func (r *resourceSyncer) ListResources() ([]runtime.Object, error) {
-	var retObjects []runtime.Object
+	retObjects := make([]runtime.Object, 0, len(r.store.List()))
```
tpantelis: This is inefficient to list it twice just to get the exact capacity.  I think 20 is a sufficient initial capacity.

tpantelis (changes requested):

pkg/syncer/resource_syncer.go
Jaanki: How about this?

```
rStoreList := r.store.List()
retObjects := make([]runtime.Object, 0, len(rStoreList))

for _, obj := range rStoreList {
```

tpantelis: yeah that makes sense although I'd shorten the var name from `rStoreList` -> `list`.

tpantelis (approved):

PR #209, opened 2021-04-16
Reconcile existing resources in broker syncer

### LINKED ISSUE #85: Reconcile existing resources in the syncer
When syncing local -> remote, on startup, the syncer should delete remote resources that no longer exist locally.

tpantelis:
Added a `Reconcile` method to the resource syncer that takes a list of resources and, if a resource does not exist in the source Store cache, it is enqueued for deletion.

The broker syncer uses this method for reconciling in both sync directions. For `RemoteToLocal`, it lists the local resources obtained from the `LocalToRemote` syncer cache. It passes the list to the `Reconcile` method on the `RemoteToLocal` syncer which only checks for remote resources, ie those with a cluster ID label set. Similarly, for `LocalToRemote`, it lists the remote resources obtained from the `RemoteToLocal` syncer cache and checks for synced local resources.

There's special consideration if the source namespace is `NamespaceAll` for reconciling `LocalToRemote`. In this case, it needs to know the namespace of the originating local resource. To handle this, when syncing, it adds a label to store the originating namespace.

Fixes #85 
Related to submariner-io/submariner#1244

**Note**: this is initially a draft b/c I want to test it with _submariner_ and _lighthouse_  but I welcome reviews in the meantime.

### DIFF
--- a/pkg/syncer/broker/syncer.go
+++ b/pkg/syncer/broker/syncer.go
@@ -293,6 +293,29 @@ func (s *Syncer) Start(stopCh <-chan struct{}) error {
 		}
 	}
 
+	lister := func(s syncer.Interface) []runtime.Object {
+		list, err := s.ListResources()
+		if err != nil {
+			klog.Errorf("Unable to reconcile - error listing resources: %v", err)
+			return nil
+		}
+
+		return list
+	}
+
+	for i := 0; i < len(s.syncers); i += 2 {
+		localSyncer := s.syncers[i]
+		remoteSyncer := s.syncers[i+1]
+
+		remoteSyncer.Reconcile(func() []runtime.Object {
+			return lister(localSyncer)
+		})
+
+		localSyncer.Reconcile(func() []runtime.Object {
+			return lister(remoteSyncer)
+		})
+	}
+
 	return nil
 }
 
--- a/pkg/syncer/broker/syncer_test.go
+++ b/pkg/syncer/broker/syncer_test.go
@@ -16,6 +16,8 @@ limitations under the License.
 package broker_test
 
 import (
+	"time"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"github.com/submariner-io/admiral/pkg/fake"
@@ -33,18 +35,20 @@ import (
 
 var _ = Describe("Broker Syncer", func() {
 	var (
-		syncer           *broker.Syncer
-		config           *broker.SyncerConfig
-		localClient      *fake.DynamicResourceClient
-		brokerClient     *fake.DynamicResourceClient
-		resource         *corev1.Pod
-		transformed      *corev1.Pod
-		initialResources []runtime.Object
-		stopCh           chan struct{}
+		syncer                 *broker.Syncer
+		config                 *broker.SyncerConfig
+		localClient            *fake.DynamicResourceClient
+		brokerClient           *fake.DynamicResourceClient
+		resource               *corev1.Pod
+		transformed            *corev1.Pod
+		initialLocalResources  []runtime.Object
+		initialBrokerResources []runtime.Object
+		stopCh                 chan struct{}
 	)
 
 	BeforeEach(func() {
-		initialResources = nil
+		initialLocalResources = nil
+		initialBrokerResources = nil
 		stopCh = make(chan struct{})
 		resource = test.NewPod("")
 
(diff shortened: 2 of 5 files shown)


pkg/syncer/broker/syncer.go
```diff
@@ -293,6 +293,29 @@ func (s *Syncer) Start(stopCh <-chan struct{}) error {
 		}
 	}
 
+	lister := func(s syncer.Interface) []runtime.Object {
+		list, err := s.ListResources()
+		if err != nil {
```
mangelajo, line 298: Should we skip the error if err is api.IsNotExist(err) ?

pkg/syncer/broker/syncer.go
```diff
@@ -293,6 +293,29 @@ func (s *Syncer) Start(stopCh <-chan struct{}) error {
...
+			klog.Errorf("Unable to reconcile - error listing resources: %v", err)
+			return nil
+		}
+
+		return list
+	}
+
+	for i := 0; i < len(s.syncers); i += 2 {
```
mangelajo, line 306: Ok, I checked that the resource syncers are ready at this point, because you wait for cache sync before exiting the syncer.Start(stopCh) function.

pkg/syncer/broker/syncer_test.go
```diff
@@ -267,6 +322,89 @@ var _ = Describe("Broker Syncer", func() {
...
+			test.SetClusterIDLabel(resource, "remote")
+			initialLocalResources = append(initialLocalResources, resource)
+
+			copy := resource.DeepCopy()
+			copy.SetNamespace(config.BrokerNamespace)
+			initialBrokerResources = append(initialBrokerResources, copy)
+		})
+
+		It("should not delete it from the local datastore on reconciliation", func() {
+			time.Sleep(100 * time.Millisecond)
```
mangelajo, line 349: I was thinking if there would be a better way for doing this, may be we can put all the sleep values under a constant at least, if we needed to bump them we could bump them all at once.

pkg/syncer/broker/syncer.go
mangelajo: Oh I see that's not possible, any error will come from WaitForCacheSync,.. ;)

mangelajo (approved):

mangelajo:
Trying on submariner repo: https://github.com/submariner-io/submariner/pull/1270
Trying on the lighthouse repo: https://github.com/submariner-io/lighthouse/pull/507
Testing in operator (not really needed): submariner-io/submariner-operator#1258

mangelajo:
Testing in lighthouse: https://github.com/submariner-io/submariner-operator/pull/1258

pkg/syncer/broker/syncer_test.go
tpantelis: This sleep isn't sensitive in that too small an interval might cause intermittent failures. I do this to give the code time to process it, such that if there's a bug and it does delete it, we can detect it. Otherwise, it may still exist b/c the code in question hasn't  executed yet.

tpantelis:
@mangelajo Thanks for the test PRs. I also intend to add unit tests cases in _submariner_ and _lighthouse_.  The latter will also likely need additional reconciliation for the stand-alone resource syncers (eg `ServiceExport` -> `ServiceImport`).

mangelajo:
Oh the lighthouse one is wrong, I threw it at the submariner-operator, I don't think we use it there. Let me try to throw one at lighthouse.

pkg/syncer/broker/syncer_test.go
mangelajo: True, I was wondering if we could wait on the reconcile process to have finished  (at that point we can check that nothing was deleted). But it's ok for now I think

mangelajo (approved):

pkg/syncer/broker/syncer_test.go
```diff
@@ -89,19 +94,25 @@ var _ = Describe("Broker Syncer", func() {
...
+		})
+
+		It("should correctly sync to the broker datastore", func() {
 			test.VerifyResource(brokerClient, resource, config.BrokerNamespace, config.LocalClusterID)
+		})
 
-			Expect(localClient.ResourceInterface.Delete(resource.GetName(), nil)).To(Succeed())
-			test.AwaitNoResource(brokerClient, resource.GetName())
+		Context("and then deleted", func() {
+			It("should be deletes from the broker datastore", func() {
```
sridhargaddam, line 108: nit: Can be updated in a different PR 
```suggestion
			It("should be deleted from the broker datastore", func() {
```

sridhargaddam (approved):
LGTM, thanks @tpantelis

pkg/syncer/resource_syncer.go
```diff
@@ -40,11 +40,15 @@ import (
 	"k8s.io/klog"
 )
 
+const OrigNamespaceLabelKey = "submariner-io/originatingNamespace"
```
vthapar, line 43: We have something similar that we use in lighthouse for namespace as well as cluster. Now that admiral tags, we should cleanup labels in lighthouse.

vthapar (approved):

pkg/syncer/broker/syncer_test.go
tpantelis: yeah we'd have to somehow provide a chan for the tests to wait on. But even that wouldn't be complete as the delete is enqueued and processed on the workqueue.

pkg/syncer/broker/syncer_test.go
tpantelis: arghhh - I missed that ☹️

pkg/syncer/resource_syncer.go
tpantelis: yes we could although admiral only tags it for a particular case but we could change that.

PR #222, opened 2021-05-19
Add update-only Federator implementation

### LINKED ISSUE #195: Add update-only Federator
The broker Federator utilizes create-or-update semantics but there are cases where a user only wants to update. Eg, when the source and destination clients are the same, we know the resource already exists so we only need to update. Also if the resource was deleted just prior, the federator could recreate it which would be problematic.

tpantelis:
There are cases where we don't want the `Federator` to create the resource if it doesn't exist, eg a resource syncer that updates the `Status` of a notified resource. In this case, if the resource was deleted while processing a create/update, we don't want it recreated. It's also more efficient to just update the Status subresource instead replacing the entire resource.

The current `Federator` implementation is in the broker package. This was refactored to the federate package into 3 `Federator` implementations: **CreateOrUpdate**, **Update**, and **UpdateStatus**.

Fixes https://github.com/submariner-io/admiral/issues/195


skitt (approved):

pkg/federate/federator_test.go
```diff
@@ -0,0 +1,379 @@
...
+	federatorNamespace string
+	targetNamespace    string
+	keepMetadataFields []string
+	dynClient          *fake.DynamicClient
+	resourceClient     *fake.DynamicResourceClient
+	restMapper         meta.RESTMapper
+	initObjs           []runtime.Object
+}
+
+func newTestDiver() *testDriver {
```
sridhargaddam:
```suggestion
func newTestDriver() *testDriver {
```

Super nit :) Maybe you meant newTestDriver()

sridhargaddam (approved):
Thanks for the PR @tpantelis

PR #226, opened 2021-05-21
Update re-queue logic in resource syncer

tpantelis:
There are cases where a transform function may return a non-nil object but still want to re-queue so the code now allows for that.


pkg/syncer/resource_syncer.go
```diff
@@ -404,16 +404,13 @@ func (r *resourceSyncer) processNextWorkItem(key, name, ns string) (bool, error)
 
 	klog.V(log.LIBTRACE).Infof("Syncer %q retrieved %sd resource %q: %#v", r.config.Name, op, resource.GetName(), resource)
```
sridhargaddam, line 405: Is %sd a typo?

sridhargaddam (approved):

pkg/syncer/resource_syncer.go
tpantelis: No - it's purposely appends a 'd' to op name.

PR #233, opened 2021-06-01
Fix update Federator

tpantelis:
K8s update requires the `ResourceVersion` field be set. Basically we need to do the same as `CreateOrUpdate` except the create so refactored that code for reuse for update-only.

Also removed the update status `Federator` - the API has some strange behavior wrt updating sub-resources. Rather then trying to debug further, we can simply use the update `Federator`.


sridhargaddam (approved):

sridhargaddam:
@mangelajo @skitt can you please take a look at this PR, TIA.

vthapar (approved):

vthapar (approved):

PR #240, opened 2021-06-23
Remove ctx parameter from For... functions

skitt:
It is never needed, the context is provided as an argument to the
various interface functions.

This is a breaking change for downstreams.

### DIFF
--- a/pkg/resource/kubernetes.go
+++ b/pkg/resource/kubernetes.go
@@ -89,7 +89,7 @@ func ForPod(client kubernetes.Interface, namespace string) Interface {
 }
 
 //nolint:dupl //false positive - lines are similar but not duplicated
-func ForService(ctx context.Context, client kubernetes.Interface, namespace string) Interface {
+func ForService(client kubernetes.Interface, namespace string) Interface {
 	return &InterfaceFuncs{
 		GetFunc: func(ctx context.Context, name string, options metav1.GetOptions) (runtime.Object, error) {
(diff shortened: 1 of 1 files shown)


tpantelis:
We can take this a step further and eliminate `Context` from the resource `Interface`.

skitt:
> We can take this a step further and eliminate `Context` from the resource `Interface`.

How so? The underlying function calls still need a context, are you suggesting we replace everything with `context.TODO()`?

tpantelis:
> > We can take this a step further and eliminate `Context` from the resource `Interface`.
> 
> How so? The underlying function calls still need a context, are you suggesting we replace everything with `context.TODO()`?

Yes. I was thinking that when the API was recently changed. Is there ever a case where it isn't `context.TODO`?

skitt:
> > > We can take this a step further and eliminate `Context` from the resource `Interface`.
> > 
> > 
> > How so? The underlying function calls still need a context, are you suggesting we replace everything with `context.TODO()`?
> 
> Yes. I was thinking that when the API was recently changed. Is there ever a case where it isn't `context.TODO`?

From reconciliation loops, it should be the context passed to the `Reconcile()` function (but I haven’t checked whether any reconciliation loops end up using these functions).

In other cases, I suspect there are circumstances where we should be using more specific contexts, _e.g._ with timeouts, or at least their cancellation functions.

tpantelis (approved):

dfarrell07 (approved):

PR #243, opened 2021-06-28
Add/update metav1.Condition utilities

tpantelis:
Taken from the submariner project:

- Changed `TryAppendCondition` to modify the `Condition` array param instead of returning a new array or nil if not changed.
- Added functions to convert `metav1.Conditions` to/from Unstructured.

### DIFF
--- a/pkg/util/condition.go
+++ b/pkg/util/condition.go
@@ -19,16 +19,17 @@ package util
 
 import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 )
 
-// TryAppendCondition appends the given Condition if it's not equal to the last Condition. Returns a new array if appended,
-// otherwise nil.
+// TryAppendCondition appends the given Condition if it's not equal to the last Condition.
 func TryAppendCondition(conditions []metav1.Condition, newCondition metav1.Condition) []metav1.Condition {
 	newCondition.LastTransitionTime = metav1.Now()
 
 	numCond := len(conditions)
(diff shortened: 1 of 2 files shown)


sridhargaddam (approved):

pkg/util/condition.go
```diff
@@ -19,21 +19,45 @@ package util
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
 )
 
-// TryAppendCondition appends the given Condition if it's not equal to the last Condition. Returns a new array if appended,
-// otherwise nil.
-func TryAppendCondition(conditions []metav1.Condition, newCondition metav1.Condition) []metav1.Condition {
+// TryAppendCondition appends the given Condition if it's not equal to the last Condition.
+func TryAppendCondition(conditions *[]metav1.Condition, newCondition metav1.Condition) {
```
skitt: Perhaps `AppendConditionIfChanged` would give a better idea of what the function does. Can’t we keep the existing function semantics, which match `append`’s? _I.e._ return the slice...

skitt (changes requested):

pkg/util/condition.go
tpantelis: Even your suggestion doesn't really convey the exact meaning (ie if what changed?). What we really want to say is `AppendConditionIfNotEqualToTheLast` but that's long. I thought `Try` reasonably conveyed that enough w/o being too long or becoming a sentence.

Originally I thought the caller would want to know if it was actually appended but that didn't turn out to be the case. I agree it makes sense to return the slice.

skitt (approved):

PR #268, opened 2021-08-20
Use SHAs for GHA versions

dfarrell07:
Per GitHub's security guidelines, GHAs should be pinned using full
length commit SHAs instead of tags.

The SHAs are of the commits currently resolved by the versions.

Even "trusted" GHAs from GitHub developers are pinned because it's
possible their repo right could be compromised and a malicious GHA
published. These core repos are not frequently substantially updated.

Submariner-internal GHAs are left pinned at devel because we want
automatic updates from Shipyard's shared tooling.

Signed-off-by: Daniel Farrell <dfarrell@redhat.com>

### DIFF
--- a/.github/workflows/codeowners.yml
+++ b/.github/workflows/codeowners.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out the repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
       - name: Delete current CODEOWNERS file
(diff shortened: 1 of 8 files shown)


dfarrell07:
YAML linting will be fixed by https://github.com/submariner-io/admiral/pull/266.

dfarrell07:
The YAML linting is actually failing, but it's not failing the job.

https://github.com/submariner-io/admiral/pull/268/checks?check_run_id=3385059564

It will be fixed by https://github.com/submariner-io/admiral/pull/266.

tpantelis (approved):

skitt (approved):

PR #278, opened 2021-10-04
Package docs: gomega, stringset

skitt:

### DIFF
--- a/pkg/stringset/stringset.go
+++ b/pkg/stringset/stringset.go
@@ -24,13 +24,31 @@ import (
 )
 
 type Interface interface {
+	// Add adds the given string to the set, if it’s not already present.
+	// Returns true if the string was added, false otherwise (it was already present).
 	Add(s string) bool
+
+	// AddAll adds all the given strings to the set.
 	AddAll(s ...string)
+
+	// Remove removes the given string from the set.
+	// Returns true if the string was present in the set, false otherwise.
 	Remove(s string) bool
+
+	// RemoveAll clears the set.
 	RemoveAll()
+
+	// Contains determines whether the set contains the given string.
 	Contains(s string) bool
+
+	// Size returns the size of the set.
 	Size() int
+
+	// Elements returns the contents of the set as a slice
 	Elements() []string
+
+	// Difference returns the difference between this set and the given object,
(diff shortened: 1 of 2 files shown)


sridhargaddam (approved):

Jaanki (approved):

pkg/stringset/stringset.go
```diff
@@ -24,13 +24,31 @@ import (
 )
 
 type Interface interface {
+	// Add adds the given string to the set, if it’s not already present.
+	// The returned value is true if the string was added, false if it was already present.
 	Add(s string) bool
+
+	// AddAll adds all the given strings to the set.
```
tpantelis, line 31:
```suggestion
	// AddAll the given strings to the set.
```

pkg/stringset/stringset.go
```diff
@@ -24,13 +24,31 @@ import (
 )
 
 type Interface interface {
+	// Add adds the given string to the set, if it’s not already present.
+	// The returned value is true if the string was added, false if it was already present.
```
tpantelis:
```suggestion
	// Add the given string to the set, if it’s not already present.
	// Returns true if the string was added, false otherwise.
```

pkg/stringset/stringset.go
```diff
@@ -24,13 +24,31 @@ import (
...
 	Add(s string) bool
+
+	// AddAll adds all the given strings to the set.
 	AddAll(s ...string)
+
+	// Remove removes the given string from the set.
+	// The return value indicates whether the string was present in the set.
```
tpantelis:
```suggestion
	// Remove the given string from the set.
	// Returns true if the string was present in the set., false otherwise.
```

tpantelis (changes requested):

pkg/stringset/stringset.go
skitt: `godoc` comments are supposed to start with the name of the element being described, treated as a noun. See https://go.dev/blog/godoc and various examples such as https://pkg.go.dev/k8s.io/apimachinery@v0.22.2/pkg/util/cache#LRUExpireCache.Remove

tpantelis (approved):

PR #281, opened 2021-10-18
Enable performance linting, fix issues

skitt:
This mostly concerns hugeParam, with fixes to avoid copying large data structures.

In many cases, we still pass by copy instead of by reference, in particular to match K8s API practice (e.g. for DeleteOptions). In such cases, the linter is disabled with // nolint:gocritic; to limit the scope of the nolint comment, it is placed inline, and where possible the affected parameter is isolated. (It doesn't seem to be possible to only disable hugeParam, and we still want hugeParam to process any additional parameter in future anyway.)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -4,7 +4,7 @@ linters-settings:
     enabled-tags:
       - diagnostic
       - opinionated
-      # - performance
+      - performance
       - style
     disabled-checks:
       - unlambda
--- a/pkg/resource/rest.go
+++ b/pkg/resource/rest.go
@@ -33,7 +33,7 @@ import (
 	"k8s.io/client-go/rest"
 )
 
-func GetAuthorizedRestConfig(apiServer, apiServerToken, caData string, tls rest.TLSClientConfig,
+func GetAuthorizedRestConfig(apiServer, apiServerToken, caData string, tls *rest.TLSClientConfig,
 	gvr schema.GroupVersionResource, namespace string) (restConfig *rest.Config, authorized bool, err error) {
 	// First try a REST config without the CA trust chain
 	restConfig, err = BuildRestConfig(apiServer, apiServerToken, "", tls)
@@ -55,7 +55,11 @@ func GetAuthorizedRestConfig(apiServer, apiServerToken, caData string, tls rest.
 	return
 }
 
-func BuildRestConfig(apiServer, apiServerToken, caData string, tls rest.TLSClientConfig) (*rest.Config, error) {
+func BuildRestConfig(apiServer, apiServerToken, caData string, tls *rest.TLSClientConfig) (*rest.Config, error) {
+	if tls == nil {
+		tls = &rest.TLSClientConfig{}
+	}
+
 	if !tls.Insecure && caData != "" {
 		caDecoded, err := base64.StdEncoding.DecodeString(caData)
 		if err != nil {
@@ -67,7 +71,7 @@ func BuildRestConfig(apiServer, apiServerToken, caData string, tls rest.TLSClien
 
 	return &rest.Config{
 		Host:            fmt.Sprintf("https://%s", apiServer),
-		TLSClientConfig: tls,
+		TLSClientConfig: *tls,
 		BearerToken:     apiServerToken,
 	}, nil
 }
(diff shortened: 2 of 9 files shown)


skitt:
Depends on https://github.com/submariner-io/releases/issues/241

pkg/resource/rest.go
```diff
@@ -33,6 +33,7 @@ import (
 	"k8s.io/client-go/rest"
 )
 
+// nolint:gocritic // rest.TLSClientConfig is big, but we need a concrete value to store in rest.Config
```
tpantelis: We can still pass a pointer here to avoid a copy and dereference when storing. But that would allow nil to be passed and we'd have to decide how to handle that. There's pros and cons to both approaches.  In this case it doesn't matter either way since this function would typically be called once.

pkg/syncer/broker/syncer.go
```diff
@@ -136,7 +136,7 @@ type Syncer struct {
 }
 
 // NewSyncer creates a Syncer that performs bi-directional syncing of resources between a local source and a central broker.
-func NewSyncer(config SyncerConfig) (*Syncer, error) {
+func NewSyncer(config *SyncerConfig) (*Syncer, error) {
```
tpantelis: This is passed by value b/c it modifies `config` although it doesn't really need to. In general, it's obviously more efficient to pass by reference but that does open up the possibility of the struct being unexpectedly mutated which might be an unwanted side effect for the  caller. In this case I think it's fine to keep this pass-by-value since it's only called once at startup so any performance hit is negligible. If we want to change it, we just have to make sure this function doesn't mutate `config`.

tpantelis (changes requested):

pkg/syncer/broker/syncer.go
skitt: Right, so since this is the only change, it probably doesn’t make sense to do this in Admiral ;-).

pkg/resource/rest.go
```diff
@@ -55,7 +55,7 @@ func GetAuthorizedRestConfig(apiServer, apiServerToken, caData string, tls rest.
 	return
 }
 
-func BuildRestConfig(apiServer, apiServerToken, caData string, tls rest.TLSClientConfig) (*rest.Config, error) {
+func BuildRestConfig(apiServer, apiServerToken, caData string, tls *rest.TLSClientConfig) (*rest.Config, error) {
```
skitt, line 58: This needs to check for `nil`.

pkg/resource/rest.go
```diff
@@ -55,7 +55,11 @@ func GetAuthorizedRestConfig(apiServer, apiServerToken, caData string, tls rest.
...
+	if tls == nil {
+		return nil, fmt.Errorf("can't build a RestConfig from nil")
+	}
```
tpantelis: We can allow nil - just use an empty struct as the caller would have done before.

```suggestion
	if tls == nil {
		tls = &rest.TLSClientConfig{}
	}
```

pkg/util/condition.go
```diff
@@ -25,15 +25,15 @@ import (
 )
 
 // TryAppendCondition appends the given Condition if it's not equal to the last Condition.
-func TryAppendCondition(conditions []metav1.Condition, newCondition metav1.Condition) []metav1.Condition {
+func TryAppendCondition(conditions []metav1.Condition, newCondition *metav1.Condition) []metav1.Condition {
```
tpantelis, line 29: Perhaps we should check for nil whether passed accidentally or not. We could log a message.

```suggestion
func TryAppendCondition(conditions []metav1.Condition, newCondition *metav1.Condition) []metav1.Condition {
    if newCondition == nil {
        klog.Warning("TryAppendCondition called with newCondition nil")
        return conditions
    }
```

tpantelis (changes requested):

tpantelis (approved):

.golangci.yml
```diff
@@ -6,6 +6,8 @@ linters-settings:
       - opinionated
       # - performance
       - style
+    enabled-checks:
+      - hugeParam
```
tpantelis: I think we should just enable `performance` on L7. The only other violation we get is:

```
pkg/syncer/broker/syncer.go:175:2: rangeValCopy: each iteration copies 168 bytes (consider pointers or indexing) (gocritic)
        for _, rc := range config.ResourceConfigs {
        ^

```

tpantelis (changes requested):

tpantelis (approved):

Jaanki (approved):

dfarrell07 (approved):

PR #284, opened 2021-10-24
Enable desired linters and document disabled ones

tpantelis:
See individual commits for details. A separate commit was created for each linter that yielded errors.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -38,6 +38,7 @@ linters:
     - gocyclo
     - godot
     - gofmt
+    - gofumpt
     - goimports
     - gosec
     - gosimple
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -48,6 +48,7 @@ linters:
     - misspell
     - nakedret
     - prealloc
+    - predeclared
     - staticcheck
     - structcheck
     - stylecheck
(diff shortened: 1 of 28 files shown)


.golangci.yml
```diff
@@ -10,55 +10,89 @@ linters-settings:
...
     - exportloopref
+    - forbidigo
     - forcetypeassert
-    # - funlen
-    # - gochecknoglobals
-    # - gochecknoinits
+    # - funlen # gocyclo is enabled which is generally a better metric than simply LOC.
+    - gci
+    # - gochecknoglobals # We don't want to forbid global variable constants
+    - gochecknoinits
```
Jaanki, line 41: Do we want this?

tpantelis: The only violations were in test code which I added an ignore.  Since this repo contains libraries, I think it makes sense for it not to have `init` side-effects.

Jaanki: Agree. Approved.

Jaanki (approved):

dfarrell07 (approved):
Nice @tpantelis, thanks! :partying_face:

skitt (approved):

PR #288, opened 2021-11-03
Hide mutexes

skitt:
This is flagged by the latest gocritic:
https://go-critic.com/overview.html#exposedSyncMutex-ref

This patch avoids exposing the mutex functions on the containing
struct.

### DIFF
--- a/pkg/fake/dynamic_client.go
+++ b/pkg/fake/dynamic_client.go
@@ -40,13 +40,12 @@ import (
 
 type DynamicClient struct {
 	*fake.FakeDynamicClient
-	sync.Mutex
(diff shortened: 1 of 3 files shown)


skitt:
Alternatively we could disable the check in gocritic, if we’d rather expose the locking/unlocking functions.

tpantelis:
> Alternatively we could disable the check in gocritic, if we’d rather expose the locking/unlocking functions.

These are fakes so not critical but I agree with the linter, ie the mutexes are meant to be internal and shouldn't be leaked.

tpantelis (approved):

dfarrell07 (approved):

PR #294, opened 2021-11-29
Add finalizer utils and test helpers

tpantelis:
Copied from the [submariner-addon](https://github.com/open-cluster-management/submariner-addon) repository.

### DIFF
--- a/pkg/finalizer/finalizer.go
+++ b/pkg/finalizer/finalizer.go
@@ -1,0 +1,85 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package finalizer
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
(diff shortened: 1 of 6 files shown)


sridhargaddam (approved):

dfarrell07 (approved):

pkg/finalizer/finalizer.go
```diff
@@ -0,0 +1,83 @@
...
+				continue
+			}
+
+			newFinalizers = append(newFinalizers, f)
+		}
+
+		objMeta.SetFinalizers(newFinalizers)
+
+		return existing, nil
+	}), "error adding finalizer %q to %q", finalizerName, objMeta.GetName())
```
skitt:
```suggestion
	}), "error removing finalizer %q from %q", finalizerName, objMeta.GetName())
```

pkg/finalizer/finalizer.go
```diff
@@ -0,0 +1,83 @@
...
+	return err == nil, errors.Wrapf(err, "error adding finalizer %q to %q", finalizerName, objMeta.GetName())
+}
+
+func Remove(ctx context.Context, client resource.Interface, obj runtime.Object, finalizerName string) error {
+	objMeta := resource.ToMeta(obj)
+	if !find(objMeta, finalizerName) {
+		return nil
+	}
+
+	return errors.Wrapf(util.Update(ctx, client, obj, func(existing runtime.Object) (runtime.Object, error) {
```
skitt: I know the error handling is different, but it would be better if the style was consistent with `Add`, *i.e.* using an `err` variable.

skitt (changes requested):

pkg/finalizer/finalizer.go
tpantelis: Done

pkg/finalizer/finalizer.go
tpantelis: Done

skitt (approved):

PR #302, opened 2021-12-17
Drop error return from BuildRestConfigFromFiles

skitt:
BuildRestConfigFromFiles always returns a nil error, drop the return
value entirely.


tpantelis (approved):
Makes sense.

dfarrell07 (approved):

PR #306, opened 2021-12-23
Adjust logr and controller-runtime dep versions

tpantelis:
...to match other submariner projects. These were recently added. Also ignore `controller-runtime` wrt **dependabot** as we do with other k8s dependencies.


tpantelis:
@davidohana

skitt (approved):
We might as well do this, but it isn’t _necessary_ — having newer versions of dependencies in descendants is fine, Go dependency resolution handles that.

Jaanki (approved):

davidohana:
LGTM. 
`logr` v.0.4.0 added the `logr.WithCallDepth()` method in an optional sub-interface but I don't think we should implement it as `controller-runtime` does not use it anyway.  Still backward compatible. https://github.com/go-logr/logr/releases/tag/v0.4.0 
BTW, if we ever bump `controller-runtime` to a recent version, we can use [zerologr](https://github.com/go-logr/zerologr) instead of the self-implementation I did for `logr.Logger`.

PR #315, opened 2022-01-31
Modify CreateAnew to check if new resource differs

tpantelis:
...from the existing one, if present. If not then don't delete and recreate. The check ignores the immutable metadata
fields and the `Status` field, if present.

### DIFF
--- a/pkg/util/create_or_update.go
+++ b/pkg/util/create_or_update.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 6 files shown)


dfarrell07 (approved):
LGTM, thanks as always @tpantelis!

pkg/util/create_or_update.go
```diff
@@ -133,16 +134,27 @@ func CreateAnew(ctx context.Context, client resource.Interface, obj runtime.Obje
...
 	name := resource.ToMeta(obj).GetName()
 
-	var created runtime.Object
+	var retObj runtime.Object
 
 	err := wait.ExponentialBackoff(backOff, func() (bool, error) {
 		var err error
 
-		created, err = client.Create(ctx, obj, createOptions)
+		retObj, err = client.Create(ctx, obj, createOptions)
```
skitt, line 142: If this returns with no error, can’t we assume that the created object is correct, and return early?

tpantelis: That's the case on L144, ie `IsAlreadyExists` returns false if err is nil.

skitt: Ah yes, my brain saw the error-handling and didn’t spot that it applied even if there was no error!

skitt (approved):

PR #322, opened 2022-02-24
Add Millisecond Precision to Zerolog Prints

davidohana:
Signed-off-by: David Ohana <david.ohana@ibm.com>

Currently its seconds precision.

### DIFF
--- a/pkg/log/kzerolog/kzerolog.go
+++ b/pkg/log/kzerolog/kzerolog.go
@@ -22,7 +22,6 @@ import (
 	"flag"
 	"fmt"
 	"os"
-	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/rs/zerolog"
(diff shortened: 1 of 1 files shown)


skitt (changes requested):
Please update your commit message to reflect the change (nanoseconds, not milliseconds) and follow the [guidelines](https://submariner.io/development/code-review/#commit-message-formatting) — in particular, “Add”, not “Added”.

Could you also explain, in your commit message, why this is desirable? Wouldn’t it be better to use our own format to get milliseconds only?

davidohana:
Initially, I preferred using a standard format e.g `RFC3339Nano`. It removes trailing zeros from the second fraction so in practice we get only milliseconds. However, it might remove too many trailing zeroes if for example the second fraction is 300 ms we will get `2022-02-24T14:21:46.3Z` and this means the timestamp length will not be constant.
That's why I've now moved to a custom format with milliseconds only and no removal of trailing zeros. 
The new custom format should also be parsable by RFC3339 parsers.

davidohana:
Demo: 
image

davidohana:
BTW commit message length enforced by the linter is too limited for long explanations :-)

dfarrell07:
> BTW commit message length enforced by the linter is too limited for long explanations :-)

You just need to wrap your lines, it doesn't preclude long explanations (https://github.com/submariner-io/shipyard/pull/730).

Update: Doesn't vim auto-wrap according to the same linter rules, if it's your commit message editor?

davidohana:
@skitt updated commit message, please re-review.

skitt (approved):

dfarrell07 (approved):

Jaanki (approved):

PR #323, opened 2022-02-25
Fix new linter errors from golangci-lint 1.44.2

dfarrell07:
See commit messages for details.

### DIFF
--- a/pkg/resource/dynamic.go
+++ b/pkg/resource/dynamic.go
@@ -54,7 +54,8 @@ func (d *dynamicType) Update(ctx context.Context, obj runtime.Object, options me
 }
 
 func (d *dynamicType) Delete(ctx context.Context, name string,
-	options metav1.DeleteOptions) error { // nolint:gocritic // Match K8s API
+	options metav1.DeleteOptions, // nolint:gocritic // Match K8s API
+) error {
 	return d.client.Delete(ctx, name, options)
 }
 
--- a/pkg/resource/interface.go
+++ b/pkg/resource/interface.go
@@ -52,6 +52,7 @@ func (i *InterfaceFuncs) Update(ctx context.Context, obj runtime.Object, options
 }
 
 func (i *InterfaceFuncs) Delete(ctx context.Context, name string,
-	options metav1.DeleteOptions) error { // nolint:gocritic // Match K8s API
+	options metav1.DeleteOptions, // nolint:gocritic // Match K8s API
+) error {
 	return i.DeleteFunc(ctx, name, options)
 }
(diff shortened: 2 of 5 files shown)


dfarrell07:
There are some more issues flagged. How should we fix these?

```
pkg/util/condition.go:56:3: type assertion must be checked (forcetypeassert)
		_ = runtime.DefaultUnstructuredConverter.FromUnstructured(rawConditions[i].(map[string]interface{}), c)
		^
pkg/util/helpers.go:85:10: type assertion must be checked (forcetypeassert)
		return value.(map[string]interface{})
		       ^
pkg/federate/create_or_update_federator.go:64:38: type assertion must be checked (forcetypeassert)
			return util.CopyImmutableMetadata(obj.(*unstructured.Unstructured), toDistribute), nil
```

tpantelis:
> There are some more issues flagged. How should we fix these?
> 
> ```
> pkg/util/condition.go:56:3: type assertion must be checked (forcetypeassert)
> 		_ = runtime.DefaultUnstructuredConverter.FromUnstructured(rawConditions[i].(map[string]interface{}), c)
> 		^
> pkg/util/helpers.go:85:10: type assertion must be checked (forcetypeassert)
> 		return value.(map[string]interface{})
> 		       ^
> pkg/federate/create_or_update_federator.go:64:38: type assertion must be checked (forcetypeassert)
> 			return util.CopyImmutableMetadata(obj.(*unstructured.Unstructured), toDistribute), nil
> ```

We can ignore these - the type assertions should never fail and, if it did, something's really weird so let it panic. We could also just disable `forcetypeassert` as it is in other repos.

skitt (approved):

nyechiel (approved):

sridhargaddam (approved):

dfarrell07:
> > There are some more issues flagged. How should we fix these?
> > ```
> > pkg/util/condition.go:56:3: type assertion must be checked (forcetypeassert)
> > 		_ = runtime.DefaultUnstructuredConverter.FromUnstructured(rawConditions[i].(map[string]interface{}), c)
> > 		^
> > pkg/util/helpers.go:85:10: type assertion must be checked (forcetypeassert)
> > 		return value.(map[string]interface{})
> > 		       ^
> > pkg/federate/create_or_update_federator.go:64:38: type assertion must be checked (forcetypeassert)
> > 			return util.CopyImmutableMetadata(obj.(*unstructured.Unstructured), toDistribute), nil
> > ```
> 
> We can ignore these - the type assertions should never fail and, if it did, something's really weird so let it panic. We could also just disable `forcetypeassert` as it is in other repos.

I sent a follow-up PR for this here: https://github.com/submariner-io/admiral/pull/324

PR #328, opened 2022-03-04
Fallback to spec values if secret is unusable

skitt:
The broker secret may be specified but unusable. In such cases, since
we provide the corresponding information in the spec, we should use
that instead of failing outright.

This reduces the impact of
https://github.com/submariner-io/submariner-operator/issues/1895.

### DIFF
--- a/pkg/syncer/broker/syncer.go
+++ b/pkg/syncer/broker/syncer.go
@@ -275,11 +275,18 @@ func createBrokerClient(config *SyncerConfig) error {
 
 		config.BrokerNamespace = spec.RemoteNamespace
 
+		// If we have a secret, try to use it
 		if spec.Secret != "" {
 			config.BrokerRestConfig, authorized, err = resource.GetAuthorizedRestConfigFromFiles(spec.APIServer,
 				filepath.Join(SecretPath(spec.Secret), "token"), filepath.Join(SecretPath(spec.Secret), "ca.crt"),
 				&rest.TLSClientConfig{Insecure: spec.Insecure}, *gvr, spec.RemoteNamespace)
-		} else {
+			if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/syncer/broker/syncer.go
```diff
@@ -275,11 +275,15 @@ func createBrokerClient(config *SyncerConfig) error {
...
+		// If we have a secret, try to use it
 		if spec.Secret != "" {
 			config.BrokerRestConfig, authorized, err = resource.GetAuthorizedRestConfigFromFiles(spec.APIServer,
 				filepath.Join(SecretPath(spec.Secret), "token"), filepath.Join(SecretPath(spec.Secret), "ca.crt"),
 				&rest.TLSClientConfig{Insecure: spec.Insecure}, *gvr, spec.RemoteNamespace)
-		} else {
+		}
+
+		// If we encountered an error, or we don't have a secret, use the values in the spec
+		if spec.Secret == "" || err != nil {
```
mkolesnik, line 289: Should we print a warning in this case instead of discarding the error out right?

Otherwise we might miss some other situations not related to the original bug

mkolesnik (changes requested):

pkg/syncer/broker/syncer.go
skitt: Unfortunately we don’t have access to a logger in this part of the code, AFAICT.

mkolesnik: I see `klog` imported and used?

skitt: Oh, thanks!

mkolesnik (approved):

sridhargaddam (approved):

skitt:
Backported as #330.

PR #342, opened 2022-04-21
Update kzerolog to use logr v1.2.0

tpantelis:
Lighthouse was updated to use K8s v1.21 so, in order to use `kzerolog`, `logr` must be bumped to v1.2.0 and the `kzerolog` code adjusted accordingly. This also requires K8s and controller-runtime to be bumped accordingly.


go.mod
```diff
@@ -5,28 +5,27 @@ go 1.13
...
+	k8s.io/apimachinery v0.23.5
+	k8s.io/client-go v0.23.5
 	k8s.io/klog v1.0.0
-	k8s.io/kube-openapi v0.0.0-20210421082810-95288971da7e // indirect
-	k8s.io/utils v0.0.0-20210305010621-2afb4311ab10 // indirect
-	sigs.k8s.io/controller-runtime v0.6.1
+	sigs.k8s.io/controller-runtime v0.11.2
+)
+
+// Pinned to kubernetes-1.21.11
```
dfarrell07, line 26: I guess pinning K8s is necessary? Wondering why. I see we do this in the operator, and pin to 19.* in the main repo.

pkg/log/kzerolog/kzerolog.go
```diff
@@ -110,11 +103,14 @@ func truncate(i interface{}, maxLen int) string {
 
 func (ctx *zeroLogContext) logEvent(evt *zerolog.Event, msg string, kvList ...interface{}) {
 	msg = truncate(ctx.prefix, maxLenLogger) + " " + msg
-	evt.Fields(kvList).CallerSkipFrame(2).Msg(msg)
+	evt.Fields(kvList).CallerSkipFrame(4).Msg(msg)
```
dfarrell07: Trying to understand this, looked at https://pkg.go.dev/github.com/rs/zerolog#Event.CallerSkipFrame but still wondering what does this change ends up doing in our context?

dfarrell07 (approved):

go.mod
tpantelis: yeah otherwise it will use  K8s 1.23 which results in build errors.

pkg/log/kzerolog/kzerolog.go
tpantelis: Well when using this with lighthouse, a value of 2 displays the file and line # in the logging lib which isn't useful, ie we want the stack frame for our calling code.  Due to changes in the updated logging libs, we now need to skip 4 frames.

tpantelis: So re: the skip frames, the first 2 call frames are for this package, eg `Info()` > `logEvent()`. In the prior logr version this was all that was needed to skip. But the newer version has a `LogSink` abstraction layer which we need to skip. Also, the **controller-runtime** log layer may add a `DelegatingLogSink` in between (depending on usage). So skip frames may be 3 or 4. I added code to dynamically calculate it.

pkg/log/kzerolog/kzerolog.go
```diff
@@ -108,13 +104,51 @@ func truncate(i interface{}, maxLen int) string {
 	return fmt.Sprintf(padFmtStr, s)
 }
 
+func (ctx *zeroLogContext) calculateSkipFrames() {
+	if ctx.skipFrames > 0 {
+		return
+	}
```
mkolesnik:
```suggestion
func (ctx *zeroLogContext) calculateSkipFrames() int {
	if ctx.skipFrames > 0 {
		return ctx.skipFrames
	}
```

I feel like it would be cleaner for callers to call `calculateSkipFrames` and just use the return value.
WDYT?

tpantelis: sure

mkolesnik (approved):

PR #385, opened 2022-07-21
Run the upgrade job on devel only

skitt:
Our upgrade test is mostly interesting for upgrades to the current
development version. On release branches, it works as long as we don't
have a release later from a later release branch; after that, it
becomes a downgrade test since we don't have a way of installing the
latest release from a given release branch.

This disables the upgrade job on all non-devel branches.

### DIFF
--- a/.github/workflows/upgrade-e2e.yml
+++ b/.github/workflows/upgrade-e2e.yml
@@ -3,6 +3,7 @@ name: Upgrade
 
 on:
   pull_request:
+    branches: [devel]
 
 jobs:
   upgrade-e2e:


.github/workflows/upgrade-e2e.yml
```diff
@@ -7,6 +7,7 @@ on:
 jobs:
   upgrade-e2e:
     name: Latest Release to Latest Version
+    if: github.base_ref == 'devel'
```
mkolesnik: Use instead?
```
on:
  pull_request:
    branches:
      - devel
```

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-branches

mkolesnik (commented):
BTW I think we could totally add some way to support getting latest release from a given prefix, the fact we dont have it now doesnt mean we cant :)

.github/workflows/upgrade-e2e.yml
mkolesnik: Just to clarify something, this will skip the job entirely, which makes more sense than the `if` since with the `if` solution the job would appear to have passed but it didnt actually do anything.

This necessitates that this job be defined as "required" only in the devel branch (which also makes sense overall), otherwise it wont let us merge if the job was skipped

mkolesnik (approved):

tpantelis (approved):

PR #528, opened 2023-01-27
Add generic function to intersect slices

tpantelis:


pkg/slices/slices_test.go
```diff
@@ -0,0 +1,56 @@
...
+				actual := slices.Intersect(s1, s2, key)
+				Expect(sets.New(actual...).Equal(sets.New(exp...))).To(BeTrue(), "Expected: %s. Actual: %s", exp, actual)
+			}
+
+			testIntersect([]string{"1", "2"}, []string{"1", "2"}, "1", "2")
+			testIntersect([]string{"1", "2"}, []string{"2", "1"}, "1", "2")
+			testIntersect([]string{"1", "2"}, []string{"1"}, "1")
+			testIntersect([]string{"3"}, []string{"1", "2", "3"}, "3")
+			testIntersect([]string{"1", "2", "3", "4"}, []string{"2", "5", "6", "3"}, "2", "3")
+		})
```
skitt, line 42: This should also have a test returning an empty set.

skitt (changes requested):

pkg/slices/slices_test.go
tpantelis: The next test case on L44 does that.

skitt: Let me clarify:

```suggestion
			testIntersect([]string{"1", "2", "3", "4"}, []string{"5", "6", "7", "8"})
		})
```

tpantelis: right  - thanks

skitt (approved):

tpantelis:
@aswinsuryan

aswinsuryan (approved):

PR #557, opened 2023-03-02
Remove our stringset implementation

skitt:
We can use k8s.io generic sets instead for unsynchronized use cases. Synchronized uses typically benefit from locks covering the full operation involving sets, rather than locks covering individual set operations, so k8s.io generic sets can also be used there with additional locking.

Depends on https://github.com/submariner-io/cloud-prepare/pull/627
Depends on https://github.com/submariner-io/lighthouse/pull/1124
Depends on https://github.com/submariner-io/submariner/pull/2324
Depends on https://github.com/submariner-io/subctl/pull/587


mkolesnik (approved):

mkolesnik:
We should probably add a "consuming projects" job for testing such changes, as the one we have in Shipyard.

tpantelis (commented):
There's another use of `stringset` [here](https://github.com/submariner-io/lighthouse/blob/d1d44e335eb54998dce95e68d3e54aadc5b16814/coredns/resolver/fake/cluster_status.go#L28). This uses the synchronized version standalone in the absence of any surrounding lock.  This code can be restructured to use an internal set and lock and expose `Add` and `Remove` methods or we can add a simple [synchronized wrapper](https://github.com/submariner-io/admiral/pull/557/commits/05d6c279d17e1818f0d861803567d36c4a1ae667) for the K8s set

skitt:
Thanks @tpantelis, can you open another PR with your synchronized set wrapper? That way we can change the users of synchronized sets to use that where appropriate, and then once that’s done delete `stringset`.

skitt:
Or we can do something like https://github.com/submariner-io/lighthouse/pull/1124.

tpantelis:
At this point we don't need a synchronized set wrapper - I'll save it off and open a PR if we ever do.

dfarrell07 (approved):

tpantelis (approved):

PR #567, opened 2023-03-13
Add generic slice functions

tpantelis:
- IndexOf
- AppendIfNotPresent
- Remove
- Union
- Equivalent

### DIFF
--- a/pkg/slices/slices.go
+++ b/pkg/slices/slices.go
@@ -18,20 +18,83 @@ limitations under the License.
 
 package slices
 
+import "reflect"
+
+// Intersect returns the intersection of two slices using the given key function.
+// The returned intersection preserves the order of elements in s2.
 func Intersect[E any, K comparable](s1, s2 []E, key func(E) K) []E {
-	m := make(map[K]bool)
-
-	for _, v := range s1 {
-		m[key(v)] = true
-	}
-
 	var r []E
 
-	for _, v := range s2 {
-		if _, ok := m[key(v)]; ok {
-			r = append(r, v)
+	m := mapFrom(s1, key)
+	for i := range s2 {
+		if _, ok := m[key(s2[i])]; ok {
+			r = append(r, s2[i])
 		}
 	}
 
 	return r
+}
+
+// IndexOf returns the index of the first occurrence of element e in s using the given key function, or -1 if not present.
+func IndexOf[E any, K comparable](s []E, e K, key func(E) K) int {
+	for i := range s {
+		if key(s[i]) == e {
+			return i
+		}
+	}
+
+	return -1
+}
+
+// AppendIfNotPresent appends element e to s if not already present using the given key function. Returns the resultant
+// slice and a bool indicating if the element was added.
+func AppendIfNotPresent[E any, K comparable](s []E, e E, key func(E) K) ([]E, bool) {
+	i := IndexOf(s, key(e), key)
+	if i >= 0 {
+		return s, false
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/slices/slices.go
```diff
@@ -18,20 +18,88 @@ limitations under the License.
...
+func AppendIfNotPresent[E any, K comparable](s []E, e E, key func(E) K) ([]E, bool) {
+	i := IndexOf(s, key(e), key)
+	if i >= 0 {
+		return s, false
+	}
+
+	return append(s, e), true
+}
+
+// Remove element e from s using the given key function. Returns the resultant slice and a bool indicating if
```
skitt:
```suggestion
// Remove the first instance of element e from s using the given key function. Returns the resultant slice and a bool indicating if
```

pkg/slices/slices.go
```diff
@@ -18,20 +18,88 @@ limitations under the License.
...
+// the element was removed.
+func Remove[E any, K comparable](s []E, v E, key func(E) K) ([]E, bool) {
```
skitt:
```suggestion
func Remove[E any, K comparable](s []E, e E, key func(E) K) ([]E, bool) {
```

pkg/slices/slices.go
```diff
@@ -18,20 +18,88 @@ limitations under the License.
...
+	i := IndexOf(s, key(v), key)
```
skitt:
```suggestion
	i := IndexOf(s, key(e), key)
```

pkg/slices/slices.go
```diff
@@ -18,20 +18,88 @@ limitations under the License.
 
 package slices
 
+// Intersect returns the intersection of two slices using the given key function.
```
skitt, line 23:
```suggestion
// Intersect returns the intersection of two slices using the given key function.
// The returned intersection preserves the order of elements in s2.
```

pkg/slices/slices.go
```diff
@@ -18,20 +18,88 @@ limitations under the License.
...
+	for i := range s2 {
+		if _, ok := m[key(s2[i])]; !ok {
+			s1 = append(s1, s2[i])
+		}
+	}
+
+	return s1
+}
+
+// Equivalent reports whether two slices have the same length and contain the same elements regardless of order.
```
skitt: What should `Equivalent()` do when presented with slices containing duplicate elements?

skitt (changes requested):

pkg/slices/slices.go
tpantelis: I guess that depends on the use case. The initial use case I have is to ignore duplicates. If we have a use case otherwise in the future we can create another function to do so

pkg/slices/slices.go
```diff
@@ -18,20 +18,84 @@ limitations under the License.
...
+		if _, ok := m[key(s2[i])]; !ok {
+			s1 = append(s1, s2[i])
+		}
+	}
+
+	return s1
+}
+
+// Equivalent reports whether two slices have the same length and contain the same elements regardless of order and
+// uniqueness.
```
skitt: Length no longer matters (if `s1` contains two elements resulting in the same key, and `s2` contains it once, they will have different lengths but be equivalent):

```suggestion
// Equivalent reports whether two slices contain the same elements regardless of order and
// uniqueness.
```

skitt (changes requested):

skitt (approved):

sridhargaddam (approved):

PR #574, opened 2023-03-23
Rename ToMeta to MustToMeta since it panics

skitt:
ToMeta panics on errors; to make that externally visible, use the Must... idiom.

ToMeta isn't used anywhere outside Admiral, so this won't require any change in other projects.


tpantelis (approved):

dfarrell07 (approved):

dfarrell07:
Good to know about the "Must" idiom, makes sense.

PR #577, opened 2023-03-24
Rename local/broker transformers for clarity

skitt:
This makes the direction of transformers and syncers explicit. The documentation is updated to match.

This will require matching changes in Submariner and Lighthouse.

Depends on https://github.com/submariner-io/releases/issues/644


tpantelis (commented):
This is an API change - let's hold off until after RC0.

dfarrell07 (approved):

dfarrell07:
The release-0.15 branch is cut, so this can go into devel.

tpantelis (approved):

PR #684, opened 2023-07-19
Add list resources by label selector method to syncer interfaces

tpantelis:
It's useful for users to query resources by selector filter rather than list all resources and filter afterwards.

The new method does not return an error, as `ListResources` does, to simplify usage. There are 2 error cases, neither of which is recoverable by the caller:
    
 - `cache.WaitForCacheSync` returns false. This means the informer was stopped so log a warning and return an empty list.
 - conversion from `Unstructured` to the resource type failed which is a programming error and shouldn't happen so use `r.convertNoError`. In fact, we really should just panic but that can be done in a separate PR.

The new method is called by `ListResources` which now can no longer return an error so error was removed from the return signature.

### DIFF
--- a/pkg/syncer/resource_syncer_test.go
+++ b/pkg/syncer/resource_syncer_test.go
@@ -872,8 +872,7 @@ func testListResources() {
 	})
 
 	It("should return all the resources", func() {
-		list, err := d.syncer.ListResources()
-		Expect(err).To(Succeed())
+		list := d.syncer.ListResources()
 		assertResourceList(list, d.resource, resource2)
 	})
 }
--- a/pkg/syncer/resource_syncer_test.go
+++ b/pkg/syncer/resource_syncer_test.go
@@ -20,6 +20,7 @@ package syncer_test
 import (
 	"context"
 	"errors"
+	"fmt"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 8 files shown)


skitt (approved):
You need to fix your IDE’s import settings ;-).

pkg/syncer/resource_syncer_test.go
```diff
@@ -870,18 +874,108 @@ func testListResources() {
...
+}
+
+func assertResourceList(actual []runtime.Object, expected ...*corev1.Pod) {
+	expSpecs := map[string]*corev1.PodSpec{}
+	for _, p := range expected {
+		expSpecs[p.Name] = &p.Spec
+	}
+
+	Expect(actual).To(HaveLen(len(expSpecs)))
+	for _, obj := range actual {
```
dfarrell07, line 968:
```
pkg/syncer/resource_syncer_test.go:972:2: ranges should only be cuddled with assignments used in the iteration (wsl)
	for _, obj := range actual {
	^
```
https://github.com/submariner-io/admiral/actions/runs/5600444498/jobs/10242837481?pr=684

dfarrell07 (approved):

tpantelis:
> You need to fix your IDE’s import settings ;-).

yeah my regular laptop is being repaired so I'm using my RH laptop and things aren't fully setup.

tpantelis:
@skitt @dfarrell07 I made some further changes which seems to have crossed with your reviews. Basically I removed error from the ListResources* methods to simplify usage.

skitt (approved):

PR #686, opened 2023-07-26
Bump to k8s.io 1.24.16 dependencies

skitt:
This is the last planned 1.24 release.


skitt:
We’ll get this automatically during the release.

PR #687, opened 2023-07-26
Handle GenerateName in util.CreateOrUpdate

tpantelis:
If `Name` is empty and `GenerateName` is set in the target resource, use the `List` method to obtain the existing resource by label selector. The motivating use case is for transforming `EndpointSlices` in **lighthouse** via a resource syncer. Instead of constructing a well-known name by concatenating various fields and risking exceeding the max length, we can use the `GenerateName` functionality. This means we need to find the target by some other means than its name. Using List by its labels makes sense although it assumes the labels uniquely identify the target resource. In order to use the `GenerateName` functionality, the caller has to ensure label uniqueness.

This requires adding a `List` method to the `resource.Interface` ([first commit](https://github.com/submariner-io/admiral/pull/687/commits/ab6549ec997088f21c7db5d8129fc226e68374a4)). In addition, add support for `GenerateName` in the fake dynamic client.

### DIFF
--- a/pkg/federate/base_federator.go
+++ b/pkg/federate/base_federator.go
@@ -46,10 +46,13 @@ func newBaseFederator(dynClient dynamic.Interface, restMapper meta.RESTMapper, t
 	keepMetadataField ...string,
 ) *baseFederator {
 	b := &baseFederator{
-		dynClient:          dynClient,
-		restMapper:         restMapper,
-		targetNamespace:    targetNamespace,
-		keepMetadataFields: map[string]bool{"name": true, "namespace": true, util.LabelsField: true, util.AnnotationsField: true},
+		dynClient:       dynClient,
+		restMapper:      restMapper,
+		targetNamespace: targetNamespace,
+		keepMetadataFields: map[string]bool{
+			"name": true, "namespace": true, "generateName": true,
+			util.LabelsField: true, util.AnnotationsField: true,
+		},
 	}
 
 	for _, field := range keepMetadataField {
--- a/pkg/util/create_or_update_test.go
+++ b/pkg/util/create_or_update_test.go
@@ -144,6 +144,10 @@ var _ = Describe("CreateOrUpdate function", func() {
 
 	createOrUpdate := func(expResult util.OperationResult) error {
 		result, err := util.CreateOrUpdate(context.TODO(), resource.ForDynamic(t.client), test.ToUnstructured(t.pod), t.mutateFn)
+		if err != nil && expResult != util.OperationResultNone {
+			return err
+		}
+
 		Expect(result).To(Equal(expResult))
 
 		return err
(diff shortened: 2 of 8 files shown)


pkg/federate/base_federator.go
```diff
@@ -65,9 +72,26 @@ func (f *baseFederator) Delete(obj runtime.Object) error {
 		return err
 	}
 
+	ctx := context.TODO()
+
+	if toDelete.GetName() == "" && toDelete.GetGenerateName() != "" {
```
skitt: The Kubernetes APIs don’t support `GenerateName` on delete operations, do we really need this?

pkg/util/create_or_update_test.go
```diff
@@ -156,6 +160,19 @@ var _ = Describe("CreateOrUpdate function", func() {
 			tests.EnsureNoActionsForResource(t.testingFake, "pods/status", "update")
 		})
 
+		Context("and GenerateName is set", func() {
+			BeforeEach(func() {
+				t.pod.Name = ""
+				t.pod.GenerateName = "name-prefix"
```
skitt: Typically,

```suggestion
				t.pod.GenerateName = "name-prefix-"
```

pkg/util/create_or_update_test.go
```diff
@@ -156,6 +160,19 @@ var _ = Describe("CreateOrUpdate function", func() {
...
+				t.pod.Labels = map[string]string{"label1": "value1", "label2": "value2"}
+			})
+
+			It("should successfully create the resource", func() {
+				Expect(createOrUpdate(util.OperationResultCreated)).To(Succeed())
+				t.verifyPod()
+			})
+		})
```
skitt, line 175: It would be worth checking that a name was generated and starts with the given prefix.

pkg/util/create_or_update_test.go
```diff
@@ -338,6 +360,23 @@ func (t *createOrUpdateTestDriver) testUpdate(doUpdate func(util.OperationResult
 			tests.EnsureNoActionsForResource(t.testingFake, "pods/status", "update")
 		})
 
+		Context("and GenerateName is set", func() {
```
skitt, line 364: I don’t think the Kubernetes APIs support `GenerateName` on update operations.

pkg/federate/base_federator.go
tpantelis: This API isn't trying to mimic/mirror the K8s delete API. The scenario here is the caller wants to delete a resource where a well-known name isn't known, ie it was generated via `GenerateName`.  The motivating use case is for transforming K8s `EndpontSlices`  to LH `EndpontSlices` via a resource syncer. Instead of constructing a well-known name by concatenating various fields and risking exceeding the max length,  we use the `GenerateName` functionality. However, on update and delete, this means we need to find the target resource via `List` by its unique labels derived from the source resource. We could do this in the caller but I thought to generalize it here and in `CreateOrUpdate` and codify the intention implicitly in the passed resource. We could codify it in the function signatures instead to make it  explicit. The other option is to have the caller obtain the target resource and not change anything in admiral.

pkg/util/create_or_update_test.go
tpantelis: yeah

pkg/util/create_or_update_test.go
tpantelis: See above comment re: delete.

pkg/federate/base_federator.go
skitt: Ah right, that makes sense. My main concern then with the generalisation is that it should apply generally — it might be sufficient to return an error if the `List` calls return more than one result...

skitt (approved):

pkg/federate/base_federator.go
tpantelis: @skitt and I discussed offline - we don't think handling `GenerateName` semantics fits here so we'll let the client handle the lookup and set the name.

pkg/util/create_or_update_test.go
tpantelis: Done

pkg/federate/base_federator.go
tpantelis: This code was removed.

vthapar (approved):

PR #688, opened 2023-07-26
Add fake client reactors to mimic real K8s behavior

tpantelis:
...for Create, Update and Delete operations to handle `GenerateName`, `ResourceVersion`, `DeletionTimestamp` et al behaviors. Also added a convenience function, `AddBasicReactors`, that adds the new reactors and the existing List, Watch and DeleteCollection reactors.

### DIFF
--- a/pkg/fake/delete_colection_reactor.go
+++ b/pkg/fake/delete_colection_reactor.go
@@ -20,29 +20,41 @@ package fake
 
 import (
 	"fmt"
+	"strings"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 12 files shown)


pkg/fake/delete_colection_reactor.go
skitt, line 1: This should be `delete_collection_reactor.go` ;-).

skitt (approved):

pkg/fake/delete_colection_reactor.go
tpantelis: oh yeah - I can fix that in another PR.

sridhargaddam (approved):

PR #690, opened 2023-07-30
Panic on resource conversion errors in resourceSyncer

tpantelis:
...instead of logging or returning the error. Such errors are unrecoverable and the result of a programming error that should be caught by unit/E2E tests (most likely a type's schema not registered). This also simplifies the code (and increases code coverage % 😄 ).


vthapar (approved):
This may cause panic in existing buggy code, but hopefully PRs updating admiral in go.mod should catch these. Would be good to update dependent projects before the next milestone/RC so as not to run into last minute errors.

tpantelis:
> This may cause panic in existing buggy code, but hopefully PRs updating admiral in go.mod should catch these. Would be good to update dependent projects before the next milestone/RC so as not to run into last minute errors.

yeah but M3 has already started. But we have no existing buggy code where conversions would fail otherwise we would see errors and functionality wouldn't work.

dfarrell07 (approved):

PR #695, opened 2023-08-01
Add RequeueResource method to the resourceSyncer

tpantelis:
This is useful to re-evaluate/retry the resource outside of the transform function, eg if the previously transformed resource is deleted out-of-band.

A specific case is with **globalnet** when a `GlobalIngressIP`, which is created via a `ServiceExport`, is deleted when the associated `Service` is deleted. If the `Service` is then re-created we need to requeue the `ServiceExport` to re-create the `GlobalIngressIP`.


pkg/syncer/resource_syncer_test.go
```diff
@@ -957,6 +958,41 @@ func testListResourcesBySelector() {
 	})
 }
 
+func testRequeueResource() {
+	d := newTestDiver(test.LocalNamespace, "", syncer.LocalToRemote)
```
sridhargaddam: nit

P.S: I know its an existing API and not added by this PR ;-) 

```suggestion
	d := newTestDriver(test.LocalNamespace, "", syncer.LocalToRemote)
```

sridhargaddam (approved):

pkg/syncer/resource_syncer_test.go
tpantelis: oh - never noticed that. I'll fix it.

dfarrell07 (approved):

PR #706, opened 2023-08-05
Increase resourceSyncer code coverage

tpantelis:
See individual commits for details.

Code coverage increased from 67% to 93.6%,

### DIFF
--- a/pkg/syncer/resource_syncer.go
+++ b/pkg/syncer/resource_syncer.go
@@ -617,21 +617,13 @@ func (r *resourceSyncer) onUpdate(oldObj, newObj interface{}) {
 }
 
 func (r *resourceSyncer) onDelete(obj interface{}) {
-	var resource *unstructured.Unstructured
-	var ok bool
-	if resource, ok = obj.(*unstructured.Unstructured); !ok {
-		tombstone, ok := obj.(cache.DeletedFinalStateUnknown)
-		if !ok {
-			r.log.Errorf(nil, "Syncer %q: could not convert object %v to DeletedFinalStateUnknown", r.config.Name, obj)
-			return
-		}
-
-		resource, ok = tombstone.Obj.(*unstructured.Unstructured)
-		if !ok {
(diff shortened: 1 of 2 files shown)


dfarrell07 (approved):
Nice!

pkg/syncer/resource_syncer.go
```diff
@@ -61,16 +61,18 @@ const (
 )
 
 func (d SyncDirection) String() string {
+	s := ""
```
skitt: The behaviour is different,

```suggestion
	s := "unknown"
```

would preserve the coverage improvement and keep the existing behaviour. I know it’s rather theoretical, but a `SyncDirection` can take any `int` value...

pkg/syncer/resource_syncer.go
```diff
@@ -82,16 +84,18 @@ const (
 )
 
 func (o Operation) String() string {
+	s := ""
```
skitt:
```suggestion
	s := "unknown"
```

pkg/syncer/resource_syncer.go
```diff
@@ -322,89 +326,99 @@ func (r *resourceSyncer) ListResources() []runtime.Object {
...
 		}
-	}()
+
+		obj := resourceUtil.MustToUnstructuredUsingScheme(resource, r.config.Scheme)
+		r.deleted.Store(key, obj)
+		r.workQueue.Enqueue(obj)
+	}
+}
+
+func (r *resourceSyncer) runIfWaitForCacheSync(defaultReturn any, run func() any) any {
```
skitt: I understand the construction, but I found “run if wait for cache sync” weird when reviewing this. `waitForCacheSyncThenRun` or `runIfCacheSynced` seem more accurate to me.

skitt (changes requested):

pkg/syncer/resource_syncer.go
tpantelis: yeah it shouldn't happen  - "unknown" isn't actually an enum so perhaps "" is more appropriate but either way...I changed it back.

skitt (approved):

PR #707, opened 2023-08-07
Add component names 

vthapar:
Adds names/names.go with submariner component names to provide consistency in component names across repos.

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -1,0 +1,28 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


sridhargaddam (approved):

pkg/version/version.go
```diff
@@ -0,0 +1,28 @@
...
+*/
+
+package version
+
+import (
+	"fmt"
+	"os"
+)
+
+func printVersion(componentName, version string) {
```
tpantelis: This needs to be exported right?

```suggestion
func Print(componentName, version string) {
```

tpantelis (changes requested):

pkg/version/version.go
sridhargaddam: Oh yeah, good catch.

vthapar: Done.

tpantelis (approved):

PR #756, opened 2023-10-06
Generify resource handling

skitt:
This simplifies various wrappers and allows the compiler to check types for us.

For now, this only addresses non-dynamic resource handlers; making the dynamic handlers generic runs into a couple of issues. The first is that the handlers are asymmetric (they take runtime.Objects as input but produce unstructured.Unstructured values as output); this can be fixed, rather nicely, by typing them and making them convert unstructured values back to their input type, but that then requires that *all* users fully specify the type. This causes problems with the federator, which makes extensive use of dynamic handlers but also expects to manage slices of syncers, which also need to be typed, but generics in Go don’t allow that.

### DIFF
--- a/pkg/resource/interface.go
+++ b/pkg/resource/interface.go
@@ -29,40 +29,40 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 )
 
-type Interface interface {
-	Get(ctx context.Context, name string, options metav1.GetOptions) (runtime.Object, error)
-	Create(ctx context.Context, obj runtime.Object, options metav1.CreateOptions) (runtime.Object, error)
-	Update(ctx context.Context, obj runtime.Object, options metav1.UpdateOptions) (runtime.Object, error)
-	UpdateStatus(ctx context.Context, obj runtime.Object, options metav1.UpdateOptions) (runtime.Object, error)
+type Interface[T runtime.Object] interface {
+	Get(ctx context.Context, name string, options metav1.GetOptions) (T, error)
+	Create(ctx context.Context, obj T, options metav1.CreateOptions) (T, error)
+	Update(ctx context.Context, obj T, options metav1.UpdateOptions) (T, error)
+	UpdateStatus(ctx context.Context, obj T, options metav1.UpdateOptions) (T, error)
 	Delete(ctx context.Context, name string, options metav1.DeleteOptions) error
-	List(ctx context.Context, options metav1.ListOptions) ([]runtime.Object, error)
+	List(ctx context.Context, options metav1.ListOptions) ([]T, error)
 }
 
-type InterfaceFuncs struct {
-	GetFunc          func(ctx context.Context, name string, options metav1.GetOptions) (runtime.Object, error)
-	CreateFunc       func(ctx context.Context, obj runtime.Object, options metav1.CreateOptions) (runtime.Object, error)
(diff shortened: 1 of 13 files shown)


tpantelis:
This is good.  I actually have this on an internal list, ie use generics where possible/appropriate.

sridhargaddam:
@skitt where can I read more about this generics?

pkg/resource/kubernetes.go
```diff
@@ -34,276 +34,122 @@ import (
...
-	}
-}
+// Controller client wrappers
 
-func ForControllerClient(client controllerClient.Client, namespace string, objType controllerClient.Object) *InterfaceFuncs {
-	return &InterfaceFuncs{
-		GetFunc: func(ctx context.Context, name string, options metav1.GetOptions) (runtime.Object, error) {
-			obj := objType.DeepCopyObject()
-			err := client.Get(ctx, types.NamespacedName{Name: name, Namespace: namespace}, obj.(controllerClient.Object))
+func ForControllerClient[T controllerClient.Object](client controllerClient.Client, namespace string, objType T) *InterfaceFuncs[T] {
```
skitt, line 133: We can drop `objType` here.

pkg/resource/kubernetes.go
```diff
@@ -316,11 +162,11 @@ func ForControllerClient(client controllerClient.Client, namespace string, objTy
 	}
 }
 
-func ForListableControllerClient(client controllerClient.Client, namespace string, objType controllerClient.Object,
+func ForListableControllerClient[T controllerClient.Object](client controllerClient.Client, namespace string, objType T,
```
skitt, line 165: We can drop `objType` and `listType` here.

pkg/resource/kubernetes.go
skitt: Actually it’s not clear we can...

pkg/resource/interface.go
```diff
@@ -71,31 +71,36 @@ func (i *InterfaceFuncs) UpdateStatus(ctx context.Context, obj runtime.Object, o
...
 	return i.ListFunc(ctx, options)
 }
 
 // DefaultUpdateStatus returns NotFound error indicating the status subresource isn't supported.
 //
 //nolint:gocritic // hugeParam - we're matching K8s API
-func DefaultUpdateStatus(_ context.Context, _ runtime.Object, _ metav1.UpdateOptions) (runtime.Object, error) {
-	return nil, apierrors.NewNotFound(schema.GroupResource{}, "")
+func DefaultUpdateStatus[T runtime.Object](_ context.Context, _ T, _ metav1.UpdateOptions) (T, error) {
+	return *new(T), apierrors.NewNotFound(schema.GroupResource{}, "")
```
tpantelis, line 89: So can't return nil (directly) since we don't know the actual type... so allocate an instance with `*new(T)` and return the zero value, which for a pointer type is nil... cool workaround.

pkg/resource/kubernetes.go
tpantelis: yeah we need to create instances of the underlying type which is problematic b/c T is actually a pointer type and there's no way to know that here hence we need an actual instance suppiled.

pkg/resource/kubernetes.go
tpantelis: We can't - we need the actual instances.

tpantelis (approved):

dfarrell07 (approved):

sridhargaddam (approved):

PR #762, opened 2023-10-16
Use the system errors package instead of github.com/pkg/errors

skitt:
... where appropriate.


tpantelis:
I like `Wrap` and `Wrapf` - saves extra if statements and having to remember %w if using `fmt.Errorf`. It's disappointing the project was archived. But that doesn't mean we can't still use it, right?

skitt:
> I like `Wrap` and `Wrapf` - saves extra if statements and having to remember %w if using `fmt.Errorf`. It's disappointing the project was archived. But that doesn't mean we can't still use it, right?

Yeah, that’s part of the discussion — the project is archived but it’s simple enough, and I don’t think there’s much risk of needing a security fix there, so it should be fine to continue using it.

I like the wrappers too; I do however feel that they sometimes lead to “write-only” code that requires too much effort to unpack, for example https://github.com/submariner-io/admiral/pull/762/files#diff-45d75e9613d7175cf26d0ec83c4c5402ce87085cafb3acddf321693311baa3a1R463-R468. But most instances are convenient and easy enough to follow (once people realise that the wrappers don’t do anything if the error is `nil`).

tpantelis:
I think we agreed we can keep using `github.com/pkg/errors`. If so, should we just close this?

skitt:
> I think we agreed we can keep using `github.com/pkg/errors`. If so, should we just close this?

I wanted to rework it to switch to `"errors"` where appropriate.

dfarrell07 (approved):

tpantelis (approved):

PR #764, opened 2023-10-19
Add LeaderElectionSupport for unit testing

tpantelis:
Refactored from the submariner project for reuse.

Related to https://github.com/submariner-io/submariner/issues/2747


skitt (commented):
Shouldn’t the various `Await` timeouts take the configured timeouts into account?

tpantelis:
> Shouldn’t the various `Await` timeouts take the configured timeouts into account?

`AwaitLeaseAcquired` could - that should take the lease duration to occur. We could use that timeout but that's really kind of internal to the leader election code and we're not trying to verify exactness in that code. But I'm fine either way.

`AwaitLeaseReleased` should happen quickly and is not dependent on any configured timeout. 

`AwaitLeaseRenewed` is called after losing the lease so, after the failure is cleared, renewal should happen after the retry period but, again, that timing seems internal to the leader election.

skitt:
Right, I’m not interested in verifying the leader election code itself; my concern was to avoid issues with tests in scenarios where the tests themselves set up specific timeouts (whether longer or short than we expect). If the timeouts involved here are internal to the leader election that’s fine.

skitt (approved):

PR #795, opened 2023-11-16
Change ForDynamic to proper generic type *Unstructured

tpantelis:
...instead of `runtime.Object`, which I think was used originally to preserve the behavior of allowing any `runtime.Object` and internally converting to `*Unstructured` but we can easily adjust users accordingly.

Some resource syncer unit tests failed b/c previously `ForDynamic` had the side effect of actually modifying the resource on `Update` and the tests did not explicitly retrieve the updated resource after `test.UpdateResource`. The fake `Federator` was also modified to internally convert to `*Unstructured` to simplify users.

### DIFF
--- a/pkg/syncer/resource_syncer_test.go
+++ b/pkg/syncer/resource_syncer_test.go
@@ -29,7 +29,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/submariner-io/admiral/pkg/federate/fake"
 	. "github.com/submariner-io/admiral/pkg/gomega"
-	"github.com/submariner-io/admiral/pkg/resource"
+	resourceutils "github.com/submariner-io/admiral/pkg/resource"
 	"github.com/submariner-io/admiral/pkg/syncer"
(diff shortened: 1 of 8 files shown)


pkg/syncer/resource_syncer_test.go
```diff
@@ -55,7 +54,7 @@ var _ = Describe("Resource Syncer", func() {
 		Context("without a local cluster ID", testRemoteToLocalWithoutLocalClusterID)
 	})
 
-	Describe("With Transform Function", testTransformFunction)
+	FDescribe("With Transform Function", testTransformFunction)
```
skitt: Typo?

tpantelis: I was doing a focus test.

skitt (approved):
The necessary changes in consumers are easy to make. Thanks!

tpantelis:
> The necessary changes in consumers are easy to make. Thanks!

yeah I changed **submariner** and **lighthouse** locally.

aswinsuryan (approved):

PR #841, opened 2024-01-31
Add helpers to stop and start metrics/profile servers

skitt:
These are used throughout the Submariner project.

### DIFF
--- a/pkg/http/server.go
+++ b/pkg/http/server.go
@@ -1,0 +1,78 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package http
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/http/pprof"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/submariner-io/admiral/pkg/log"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var logger = log.Logger{Logger: logf.Log}
+
+type EndpointType int
+
+const (
+	Metrics EndpointType = 1 << iota
+	Profile
+	all = Metrics | Profile
+)
+
+type StopFunc func()
+
+// StartServer starts an HTTP server providing access to Prometheus metrics
+// and/or profiling information.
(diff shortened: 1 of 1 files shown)


pkg/util/helpers.go
```diff
@@ -171,3 +176,41 @@ func AddCertificateErrorHandler(fatal bool) {
 		// The generic handler has already logged the error, no need to repeat if we don't want extra detail
 	})
 }
+
+// StartMetricsAndProfileServer starts an HTTP server providing access to Prometheus metrics
+// and/or profiling information.
+// The returned server should be shut down on process exit, using StopMetricsAndProfileServer().
+func StartMetricsAndProfileServer(metrics, profile bool, port int) *http.Server {
```
tpantelis: Users don't actually need the `http.Server ` other than to close it so how about returning a function that closes it instead. This obviates the need for a separate Stop* function.

```suggestion
type StopFunc func()

func StartHTTPServer(metrics, profile bool, port int) StopFunc{
     ...
    return func() {
        if err := srv.Shutdown(context.TODO()); err != nil {
		logger.Errorf(err, "Error shutting down metrics/profile HTTP server")
	}
    }
}
```

tpantelis: We could put this in an `http` package:` http.StartServer`. I've thought about moving the functions in the `util` package to other, suitably-named packages.

tpantelis (changes requested):

pkg/http/server.go
```diff
@@ -0,0 +1,68 @@
...
+
+var logger = log.Logger{Logger: logf.Log}
+
+type StopFunc func()
+
+// StartMetricsAndProfileServer starts an HTTP server providing access to Prometheus metrics
+// and/or profiling information.
+// The returned function should be called to shut down the server, typically at process exit
+// using defer in the main() function.
+func StartMetricsAndProfileServer(metrics, profile bool, port int) StopFunc {
```
tpantelis: nit - "MetricsAndProfile" are options (passed as params) and I don't think need/should to be part of the name. Also technically it should be "MetricsAndOrProfile" 😄  

```suggestion
func StartServer(metrics, profile bool, port int) StopFunc {
```

tpantelis (changes requested):

pkg/http/server.go
skitt: I actually didn’t like the use of bools, so let’s avoid those altogether.

tpantelis (approved):

pkg/http/server.go
```diff
@@ -0,0 +1,78 @@
...
+// and/or profiling information.
+// endpoints specifies the endpoints to provide; it can be METRICS, PROFILE, or both (ored).
+// port specifies the TCP port on which the server listens.
+// The returned function should be called to shut down the server, typically at process exit
+// using defer in the main() function.
+func StartServer(endpoints EndpointType, port int) StopFunc {
+	if endpoints&all == 0 {
+		return func() {}
+	}
+
```
tpantelis, line 56: Perhaps not defining any endpoints should be an error:

```suggestion
	if endpoints == 0 {
		panic("An endpoints must be specified")
	}
```

skitt: I’d rather not have that enforced in Admiral — if we make endpoint setup user-configurable in future, it would mean more boilerplate code in programs using this.

vthapar (approved):

dfarrell07:
Do we want to backport this?

PR #891, opened 2024-04-16
Fix data race due to trimming ManagedFields in resource syncer

tpantelis:
This was observed in a unit test elsewhere that configures a resync period. On re-sync, the K8s `DeltaFIFO` retrieves every object from the cache store and re-queues them. It also invokes the transform function which results in a data race when the resource syncer tries to nil the `ManagedFields`. This is because the object instance is the same as that stored in the cache which was previously accessed by another thread. So mutating it without the protection of a lock is unsafe. This is really an issue with the `DeltaFIFO` - it is documented that the "_TransformFunc sees the object before any other actor, and it is now safe to mutate the object in place instead of making a copy_" however this is not the case on a re-sync. The `DeltaFIFO` should either elide the `TransformFunc` on re-sync or make a copy of the object retrieved from the cache store.

As a workaround, the resource syncer should only set `ManagedFields` to nil if it is non-nil, which will be the case a a re-sync. Added a unit test to cover this case.

Also the object passed to the `TransformFunc` could be a `DeletedFinalStateUnknown` so we need to handle that as well.

### DIFF
--- a/pkg/syncer/resource_syncer.go
+++ b/pkg/syncer/resource_syncer.go
@@ -257,7 +257,11 @@ func NewResourceSyncer(config *ResourceSyncerConfig) (Interface, error) {
 		UpdateFunc: syncer.onUpdate,
 		DeleteFunc: syncer.onDelete,
 	}, func(obj interface{}) (interface{}, error) {
-		resourceUtil.MustToMeta(obj).SetManagedFields(nil)
+		objMeta, err := meta.Accessor(obj)
+		if err == nil && len(objMeta.GetManagedFields()) > 0 {
+			objMeta.SetManagedFields(nil)
+		}
+
 		return obj, nil
 	})
 
--- a/pkg/syncer/resource_syncer_test.go
+++ b/pkg/syncer/resource_syncer_test.go
@@ -45,6 +45,7 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/dynamic"
 	fakeClient "k8s.io/client-go/dynamic/fake"
+	"k8s.io/utils/ptr"
 )
 
 var _ = Describe("Resource Syncer", func() {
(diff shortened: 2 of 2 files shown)


pkg/syncer/resource_syncer_test.go
```diff
@@ -1183,6 +1185,37 @@ func testRequeueResource() {
...
+	})
+
+	It("should remove ManagedFields from created resources", func() {
+		obj, exists, err := d.syncer.GetResource(d.resource.Name, d.resource.Namespace)
+		Expect(err).To(Succeed())
+		Expect(exists).To(BeTrue())
+		Expect(resourceutils.MustToMeta(obj).GetManagedFields()).Should(BeNil())
+
+		// Sleep a little so a re-sync occurs and doesn't cause a data race.
+		time.Sleep(200)
```
skitt, line 1215: Shouldn’t something be checked after the sleep?

skitt:
Has this been reported upstream?

pkg/syncer/resource_syncer_test.go
tpantelis: There's nothing we can check here but Ginkgo will fail with a data race if it occurs. So we're just trying to trigger a data race here.

tpantelis:
> Has this been reported upstream?

https://github.com/kubernetes/kubernetes/issues/124337

skitt (approved):

yboaron (approved):

pkg/syncer/resource_syncer.go
```diff
@@ -257,7 +257,11 @@ func NewResourceSyncer(config *ResourceSyncerConfig) (Interface, error) {
 		UpdateFunc: syncer.onUpdate,
 		DeleteFunc: syncer.onDelete,
 	}, func(obj interface{}) (interface{}, error) {
-		resourceUtil.MustToMeta(obj).SetManagedFields(nil)
+		objMeta, err := meta.Accessor(obj)
```
skitt, line 260: On the face of it, this changes the behaviour if the given object doesn’t have the expected metadata fields — the previous version would panic, the new version ignores the error. This is unlikely anyway, and if there is no metadata then there is nothing to update so ignoring the error is fine, but it would be worth adding a comment to mention that errors aren’t returned on purpose.

PR #893, opened 2024-04-17
Add reusable TrimManagedFields transform function

tpantelis:


pkg/resource/util.go
```diff
@@ -125,3 +125,12 @@ func ToJSON(o any) string {
 	out, _ := json.MarshalIndent(o, "", "  ")
 	return string(out)
 }
+
+func TrimManagedFields(obj interface{}) (interface{}, error) {
+	objMeta, err := meta.Accessor(obj)
```
skitt, line 132: On the face of it,  #891 changed the behaviour if the given object doesn’t have the expected metadata fields — the previous version would panic, the new version (which this is turning into a function) ignores the error. This is unlikely anyway, and if there is no metadata then there is nothing to update so ignoring the error is fine, but it would be worth adding a comment to mention that errors aren’t returned on purpose.

skitt (changes requested):

pkg/resource/util.go
tpantelis: I could add a comment that this is intended to be a `TransformFunc` and ignores non-MetaData objects.

skitt (approved):

dfarrell07 (approved):

PR #895, opened 2024-04-24
Add support for resource syncer to use a shared informer

tpantelis:
Some users may want to configure a shared informer rather than use the internal dedicated informer for efficiency if there's multiple consumers of a resource type.


tpantelis:
Ping @aswinsuryan @vthapar @dfarrell07

dfarrell07 (approved):

aswinsuryan (approved):

PR #905, opened 2024-05-04
Add IsMissingNamespaceErr function

tpantelis:
Checks if the error indicates the namespace for a resource is missing when trying to create it. In this case the error is NotFound and the status details reflects the "namespaces" Kind. For example:

```
{
  "ErrStatus": {
    "status": "Failure",
    "message": "namespaces \"foo\" not found",
    "reason": "NotFound",
    "details": {
      "name": "foo",
      "kind": "namespaces"
    },
    "code": 404
  }
}

```

### DIFF
--- a/pkg/resource/errors.go
+++ b/pkg/resource/errors.go
@@ -42,3 +42,16 @@ func IsNotFoundErr(err error) bool {
 
 	return false
 }
+
+func IsMissingNamespaceErr(err error) (bool, string) {
+	if !apierrors.IsNotFound(err) {
+		return false, ""
+	}
+
+	var status apierrors.APIStatus
+	_ = errors.As(err, &status)
+
+	d := status.Status().Details
+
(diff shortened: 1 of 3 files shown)


pkg/resource/errors.go
```diff
@@ -42,3 +42,16 @@ func IsNotFoundErr(err error) bool {
 
 	return false
 }
+
+func IsMissingNamespaceErr(err error) (bool, string) {
```
skitt, line 46: We should disabled `unnamedResult` in `gocritic` — it’s disabled in all the other projects.

skitt (approved):

pkg/resource/errors.go
tpantelis: yeah I missed that.

aswinsuryan (approved):

PR #912, opened 2024-05-09
Adjust to golangci-lint 1.58.0

tpantelis:


aswinsuryan (approved):

.golangci.yml
```diff
@@ -169,5 +169,6 @@ issues:
         - contextcheck
         - errname
         - gochecknoinits
-        - goerr113
+        - init
```
skitt: Why is this being added?

tpantelis: To workaround an issue with `gochecknoinits` : https://github.com/golangci/golangci-lint/issues/4697.

tpantelis: Perhaps I should comment this 😄

skitt: Could you add a comment to that effect, to avoid head-scratching if this causes errors once golangci-lint is fixed?

skitt (changes requested):

.golangci.yml
tpantelis: just did

skitt: Hah I hadn’t seen your comment!

skitt (approved):

dfarrell07 (approved):

PR #979, opened 2024-08-23
EnsureValidName should strip leading or trailing dash

tpantelis:
`EnsureValidName` converts all invalid chars to a dash ('-') but that could leave a leading or trailing dash which is invalid so just strip a leading or trailing dash.

### DIFF
--- a/pkg/resource/util.go
+++ b/pkg/resource/util.go
@@ -111,7 +111,7 @@ func MustToMeta(obj interface{}) metav1.Object {
 func EnsureValidName(name string) string {
 	// K8s only allows lower case alphanumeric characters, '-' or '.'. Regex used for validation is
 	// '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'
-	return strings.Map(func(c rune) rune {
+	name = strings.Map(func(c rune) rune {
 		c = unicode.ToLower(c)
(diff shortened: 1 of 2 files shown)


pkg/resource/util.go
```diff
@@ -111,14 +111,25 @@ func MustToMeta(obj interface{}) metav1.Object {
...
 	}, name)
+
+	if name[0] == '-' {
+		name = name[1:]
+	}
+
+	nameLen := len(name)
+	if name[nameLen-1] == '-' {
+		name = name[:nameLen-1]
+	}
```
skitt: There could be multiple dashes at the start or end of the string.

```suggestion
	name = strings.Trim(name, "-")
```

The resulting name could even be empty in pathological cases — this is unlikely in the Submariner codebase so it might not be worth handling.

skitt (changes requested):

pkg/resource/util.go
tpantelis: yeah this is used to sanitize host names and cluster IDs. The former can only start with a letter or digit. For the latter, `subctl` ensures the validity of cluster IDs and ACM uses cluster IDs as resource names which also must start with a letter or digit. So it's only the helm charts that doesn't validate. Regardless it is highly unlikely someone would specify a cluster ID with no letters or digits (unless maybe for DOS attack). But another possibility, though unlikely, is that, after sanitizing, two cluster IDs are now equal.

skitt (approved):

vthapar (approved):

PR #1052, opened 2025-01-24
Implement priority queueing in the resource syncer

tpantelis:
All pre-existing resources are queued by the informer on startup. With many resources in the hundreds or thousands it can take significant time to process them all and the work queue rate limiter can cause further significant delays. This can block newly created resources or existing resource updates for potentially minutes. Most of the time a pod restart is fairly quick so most resources didn't likely change so it makes sense to give newly created/updated resources higher priority and process them in a more timely manner.  

The **client-go** work queue allows the backend `Queue` implementation to be configurable. This PR adds an implementation to the **admiral** `workqueue` module that is backed by a priority queue that is used in conjunction with the `container/heap` module and implements `heap.Interface`. A new `EnqueueWithPriority` method was added to the **admiral** `workqueue.Interface` that allows the caller to specify the priority and whether or not to enqueue with rate limiting. The existing `Enqueue` method was modified to enqueue with normal priority 0. 

The resource syncer was modified to use the `EnqueueWithPriority` method to prioritize newly created/updated resources over pre-existing resources on startup. The `cache.ResourceEventHandler'`s `OnAdd` method has an `isInInitialList` param that indicates if the obj was pre-existing in the initial listing retrieved from the API server. If `isInInitialList` is true then enqueue with low priority otherwise use normal priority (`Enqueue` method).

See commits for more details.

Related to https://github.com/submariner-io/lighthouse/issues/1706

### DIFF
--- a/pkg/workqueue/config.go
+++ b/pkg/workqueue/config.go
@@ -1,0 +1,105 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package workqueue
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+)
+
+const (
+	ItemRateLimiterBaseDelayKey     = "item-rate-limiter-base-delay"
+	ItemRateLimiterMaxDelayKey      = "item-rate-limiter-max-delay"
+	OverallRateLimiterMaxDelayKey   = "overall-rate-limiter-max-delay"
+	BucketRateLimiterItemsPerSecKey = "bucket-rate-limiter-items-per-sec"
+	BucketRateLimiterMaxBurstKey    = "bucket-rate-limiter-max-burst"
+)
+
+type Config struct {
+	ItemRateLimiterBaseDelay     time.Duration
+	ItemRateLimiterMaxDelay      time.Duration
+	OverallRateLimiterMaxDelay   time.Duration
+	BucketRateLimiterItemsPerSec int
+	BucketRateLimiterMaxBurst    int
+}
+
+func DefaultConfig() Config {
+	return Config{
+		ItemRateLimiterBaseDelay:     time.Millisecond,
+		ItemRateLimiterMaxDelay:      30 * time.Second,
+		OverallRateLimiterMaxDelay:   5 * time.Minute,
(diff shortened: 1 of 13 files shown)


pkg/workqueue/queue.go
```diff
@@ -69,19 +78,36 @@ func NewWithConfig(name string, config Config) Interface {
...
+						Queue: &priorityWorkQueue[string]{priorityQueue: priorityQueue},
+					}),
+				}),
 			}),
 		name: name,
 	}
 }
 
-func (q *queueType) Enqueue(obj interface{}) {
+func (q *queueType) enqueue(obj interface{}, priority int) {
```
skitt: Could the priority use a specific type? It would make it more obvious that this expects one of the two constants, and not an arbitrary priority. (This isn’t all that important since the method is private.)

tpantelis: We could but like you said it's private. I only defined and exposed two priorities (via specific functions) at least for now. Perhaps we'll have a future need for users to specify arbitrary priorities in which case a specific type wouldn't make sense. I figure we can deal with that if/when. For now I kept it simple with "low" and "normal".

tpantelis: Another approach is to move the low and normal priority concept internally to the resource syncer and have the work queue take an arbitrary priority, ie change `EnqueueLowPriority` to `EnqueueWithPriority(...int priority)`.

pkg/workqueue/priority_queue.go
```diff
@@ -0,0 +1,132 @@
...
+		items:   []itemType{},
+		indices: map[any]int{},
+	}
+}
+
+func (p *PriorityQueue) Len() int {
+	return len(p.items)
+}
+
+func (p *PriorityQueue) Less(i, j int) bool {
```
yboaron, line 55: QQ: does lower number mean higher priority ?

tpantelis: No. Thus function determines if the element at index i should be moved before element j in the queue so that it is de-queued first. The naming of the function is a bit confusing.

yboaron (approved):

pkg/workqueue/queue.go
tpantelis: I changed the method to `EnqueueWithOpts` and also added a `RateLimiting` flag (to mirror `AddWithOpts` in the new controller-runtime [PriorityQueue](https://github.com/kubernetes-sigs/controller-runtime/blob/626b2f36e1b6ef8daa1bec519a282f49618f0d71/pkg/controller/priorityqueue/priorityqueue.go#L30)). Looking at the K8s controller-runtime code, they do not enqueue items from the initial listing on startup with rate limiting which makes sense b/c it already knows it's a one-time burst.  I did the same in the resource syncer.

pkg/workqueue/config.go
```diff
@@ -0,0 +1,105 @@
...
+	if c == nil {
+		return DefaultConfig()
+	}
+
+	return *c
+}
+
+func ConfigFromConfigMap(configMap *corev1.ConfigMap, keyPrefix string, defaultConfig *Config) *Config {
+	if configMap == nil {
+		return defaultConfig
```
skitt, line 67: This is _really_ minor, but perhaps

```suggestion
		return DefaultConfigIfNil(defaultConfig)
```

tpantelis: But we've already checked for nil on L66 so we don't need to call `DefaultConfigIfNil`.

tpantelis: I didn't do that b/c I assumed the caller probably wants nil returned if they passed nil `defaultConfig`.

skitt (approved):

PR #1061, opened 2025-02-05
Use controller-runtime PriorityQueue

tpantelis:
The 0.20.1 version includes a priority queue implementation so we can use that in lieu of our own.


tpantelis:
The unit test failures highlight that the priority work queue does not shut down as we expect and I've seen a code comment which indicates it's really designed to work with other components of controller-runtime so I think we should stay with what we have to be safe.

PR #1080, opened 2025-03-08
Add function to get a ConfigMap or return default instance

tpantelis:
If the `ConfigMap` does not exist then return an instance with an empty `Data` map. This is a convenience function for callers that don't care if the `ConfigMap` does not exist.

### DIFF
--- a/pkg/configmap/config_map.go
+++ b/pkg/configmap/config_map.go
@@ -1,0 +1,50 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


pkg/configmap/config_map.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"context"
+
+	"github.com/pkg/errors"
+	"github.com/submariner-io/admiral/pkg/resource"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const Global = "submariner-global"
```
yboaron, line 31: Maybe we should add a comment to explain what "submariner-global" CM is

yboaron (approved):

yboaron (approved):

skitt (approved):

PR #1101, opened 2025-04-07
Add Context param to AwaitStopped

tpantelis:
...and return an error if the `Context` times out or is canceled. This lets the caller control the behavior if shut down is delayed.

### DIFF
--- a/pkg/syncer/syncer.go
+++ b/pkg/syncer/syncer.go
@@ -19,13 +19,15 @@ limitations under the License.
 package syncer
 
 import (
+	"context"
+
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
 type Interface interface {
 	Start(stopCh <-chan struct{}) error
-	AwaitStopped()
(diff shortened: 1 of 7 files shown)


pkg/syncer/syncer.go
```diff
@@ -19,13 +19,15 @@ limitations under the License.
...
+
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
 type Interface interface {
 	Start(stopCh <-chan struct{}) error
-	AwaitStopped()
+	AwaitStopped(ctx context.Context) error
 	GetResource(name, namespace string) (runtime.Object, bool, error)
```
skitt, line 31: Not a blocker for this PR, but shouldn’t all the syncer operations accept a context and handle cancellation?

skitt (approved):

pkg/syncer/syncer.go
tpantelis: Perhaps but this and others are  non-blocking and internal invocations don't take a context.

vthapar (approved):

PR #1107, opened 2025-04-08
[0.16] Bump oauth2 0.7.0 to 0.27.0 for CVE-2025-22868

dfarrell07:
Change generated by

  go get golang.org/x/oauth2@v0.27.0
  go mod tidy
  <manually remove go toolchain directive>


skitt:
For future reference, I reckon it would be better to make these changes (where possible, as in this instance) in Shipyard only, and rely on the release process to update the dependencies everywhere — that would result in a lot less work for the initial submitter and all the reviewers. Projects with separate tools modules might still need a separate bump, but CVEs have less impact there.

skitt (approved):

tpantelis:
> For future reference, I reckon it would be better to make these changes (where possible, as in this instance) in Shipyard only, and rely on the release process to update the dependencies everywhere — that would result in a lot less work for the initial submitter and all the reviewers. Projects with separate tools modules might still need a separate bump, but CVEs have less impact there.

yeah I suggested that too over slack.

tpantelis (approved):

PR #1110, opened 2025-04-08
Modernise code using maps/slices functions, int ranges, and any

skitt:
See individual commits.

### DIFF
--- a/pkg/command/fake/command.go
+++ b/pkg/command/fake/command.go
@@ -60,7 +60,7 @@ type pipeReader struct {
 }
 
 type commandOutputInfo struct {
-	pathMatcher  interface{}
+	pathMatcher  any
 	expectedArgs []any
 	output       string
 	err          error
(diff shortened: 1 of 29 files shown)


dfarrell07:
Do we want to backport this? Talking with team, it seems like refactoring that we wouldn't normally backport but since it touches so many lines it might make other backport easier if it's also backported.

tpantelis (approved):

skitt:
> Do we want to backport this? Talking with team, it seems like refactoring that we wouldn't normally backport but since it touches so many lines it might make other backport easier if it's also backported.

Perhaps I can drop the `any` commit instead, that’s the intrusive one and it’s really just style.

tpantelis:
> Perhaps I can drop the `any` commit instead, that’s the intrusive one and it’s really just style.

I think it's fine. We can't be afraid of changing/improving code b/c it may make backporting harder. Plus most of it's code that rarely changes anyway let alone needing to backport a change.

dfarrell07 (approved):

PR #1112, opened 2025-04-10
Add default timeout in AwaitStopped and ShutDownWithDrain

tpantelis:
...in case the caller doesn't set a deadline in the `Context`. This preserves the previous behavior of `ShutDownWithDrain` which did set a deadline internally.

### DIFF
--- a/pkg/syncer/resource_syncer.go
+++ b/pkg/syncer/resource_syncer.go
@@ -431,6 +431,9 @@ func (r *resourceSyncer) shutDownWorkQueue() {
 }
 
 func (r *resourceSyncer) AwaitStopped(ctx context.Context) error {
+	ctx, cancel := context.WithTimeout(ctx, r.config.DrainWorkQueueTimeout)
+	defer cancel()
+
 	select {
 	case _, closed := <-r.stopped:
 		if closed {
--- a/pkg/workqueue/queue.go
+++ b/pkg/workqueue/queue.go
@@ -22,6 +22,7 @@ package workqueue
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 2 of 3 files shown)


aswinsuryan (approved):

skitt (approved):
I’m guessing it doesn’t matter that this overwrites any deadline set in the provided context.

tpantelis:
> I’m guessing it doesn’t matter that this overwrites any deadline set in the provided context.

It doesn't overwrite it: "_WithDeadline returns a copy of the parent context with the deadline adjusted to be no later than d. If the parent's deadline is already earlier than d, WithDeadline(parent, d) is semantically equivalent to parent._ ".

skitt:
> > I’m guessing it doesn’t matter that this overwrites any deadline set in the provided context.
> 
> It doesn't overwrite it: "_WithDeadline returns a copy of the parent context with the deadline adjusted to be no later than d. If the parent's deadline is already earlier than d, WithDeadline(parent, d) is semantically equivalent to parent._ ".

So it does, thanks! I checked the docs for `WithTimeout` but didn’t scroll up to `WithDeadline` :-(.

tpantelis:
> > > I’m guessing it doesn’t matter that this overwrites any deadline set in the provided context.
> > 
> > 
> > It doesn't overwrite it: "_WithDeadline returns a copy of the parent context with the deadline adjusted to be no later than d. If the parent's deadline is already earlier than d, WithDeadline(parent, d) is semantically equivalent to parent._ ".
> 
> So it does, thanks! I checked the docs for `WithTimeout` but didn’t scroll up to `WithDeadline` :-(.

But if the parent timeout is longer then it will override. I really just want  a default timeout so it doesn't potentially block indefinitely. Perhaps we should set a timeout if not already set in the parent. I can push a followup.

PR #1119, opened 2025-04-29
Make log verbosity configurable in the work queue

tpantelis:
...and add more logging.

### DIFF
--- a/pkg/log/logger.go
+++ b/pkg/log/logger.go
@@ -87,3 +87,14 @@ func (l Logger) V(level int) Logger {
 	l.Logger = l.Logger.V(level)
 	return l
 }
+
+func (l Logger) SetMaxVerbosity(v int) {
+	type configurableLogSink interface {
+		SetMaxVerbosity(v int)
+	}
+
+	c, ok := l.Logger.GetSink().(configurableLogSink)
+	if ok {
+		c.SetMaxVerbosity(v)
+	}
+}
--- a/pkg/log/kzerolog/kzerolog.go
+++ b/pkg/log/kzerolog/kzerolog.go
@@ -230,3 +230,7 @@ func (ctx *zeroLogContext) WithValues(kvList ...any) logr.LogSink {
 
 	return &subCtx
 }
+
+func (ctx *zeroLogContext) SetMaxVerbosity(v int) {
+	ctx.maxVerbosity = v
+}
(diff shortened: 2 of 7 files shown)


dfarrell07:
Do you want to merge this or is it for debugging/testing?

dfarrell07 (approved):

aswinsuryan (approved):

pkg/log/logger.go
```diff
@@ -87,3 +91,10 @@ func (l Logger) V(level int) Logger {
 	l.Logger = l.Logger.V(level)
 	return l
 }
+
+func (l Logger) SetMaxVerbosity(v int) {
+	c, ok := l.Logger.GetSink().(ConfigurableLogSink)
+	if ok {
+		c.SetMaxVerbosity(v)
+	}
+}
```
skitt: Since `ConfigurableLogSink` isn’t used anywhere else (and shouldn’t be, AFAICT), how about

```suggestion
func (l Logger) SetMaxVerbosity(v int) {
	type configurableLogSink interface {
		SetMaxVerbosity(v int)
	}

	c, ok := l.Logger.GetSink().(configurableLogSink)
	if ok {
		c.SetMaxVerbosity(v)
	}
}
```

skitt (changes requested):

pkg/log/logger.go
tpantelis: It's public so it's visible to implementors, currently only `zeroLogContext`. I changed it to `configurableLogSink` and the compiler still recognizes `zeroLogContext` as an implementation but it seems odd.

skitt: That’s perfectly fine — Go is close to duck typing. Go interfaces aren’t so much an implementation contract as they are a _type usage_ contract, and in many cases they’re relevant for code _using_ types, not implementing them. See https://go.dev/wiki/CodeReviewComments#interfaces for background. In this instance, the interface seems like a proxy for a method presence check — that is to say, I understand the intent of the code to be “if `l.logger.GetSink()` has a `SetMaxVerbosity` method taking an `int` parameter, call it”. The way to represent that in Go is with a private, usage-side interface.

tpantelis: OK.  I guess the benefit to having it public is to ensure the desired interface is implemented, ie

`var _ log.ConfigurableLogSink = &zeroLogContext{}`

But I'm fine either way - I don't think it really matters.

skitt (approved):

PR #1121, opened 2025-04-30
Make log verbosity configurable in the work queue

tpantelis:
...and add more logging.

### DIFF
--- a/pkg/log/logger.go
+++ b/pkg/log/logger.go
@@ -84,3 +84,10 @@ func (l Logger) V(level int) Logger {
 	l.Logger = l.Logger.V(level)
 	return l
 }
+
+func (l Logger) SetMaxVerbosity(v int) {
+	c, ok := l.Logger.GetSink().(interface{ SetMaxVerbosity(v int) })
+	if ok {
+		c.SetMaxVerbosity(v)
+	}
+}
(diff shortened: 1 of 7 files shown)


dfarrell07 (approved):

pkg/log/logger.go
```diff
@@ -30,6 +30,10 @@ const (
 	FatalKey   = "FATAL"
 )
 
+type ConfigurableLogSink interface {
+	SetMaxVerbosity(v int)
+}
+
```
skitt:
```suggestion
```

pkg/log/logger.go
```diff
@@ -84,3 +88,10 @@ func (l Logger) V(level int) Logger {
 	l.Logger = l.Logger.V(level)
 	return l
 }
+
+func (l Logger) SetMaxVerbosity(v int) {
+	c, ok := l.Logger.GetSink().(ConfigurableLogSink)
```
skitt: Same as `devel`:

```suggestion
	c, ok := l.Logger.GetSink().(interface{ SetMaxVerbosity(v int) })
```

skitt (changes requested):

skitt (approved):

PR #1128, opened 2025-05-06
Replace configurableLogSink with an anonymous interface

skitt:
This is a private check for a method, it doesn't need to be named at all.

### DIFF
--- a/pkg/log/logger.go
+++ b/pkg/log/logger.go
@@ -89,11 +89,7 @@ func (l Logger) V(level int) Logger {
 }
 
 func (l Logger) SetMaxVerbosity(v int) {
-	type configurableLogSink interface {
-		SetMaxVerbosity(v int)
-	}
-
-	c, ok := l.Logger.GetSink().(configurableLogSink)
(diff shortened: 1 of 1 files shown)


pkg/log/logger.go
```diff
@@ -89,11 +89,7 @@ func (l Logger) V(level int) Logger {
 }
 
 func (l Logger) SetMaxVerbosity(v int) {
-	type configurableLogSink interface {
-		SetMaxVerbosity(v int)
-	}
-
-	c, ok := l.Logger.GetSink().(configurableLogSink)
+	c, ok := l.Logger.GetSink().(interface{ SetMaxVerbosity(int) })
```
tpantelis: Neat simplification.

tpantelis (approved):

pkg/log/logger.go
skitt: Thanks, I hadn’t thought of it earlier!

dfarrell07 (approved):

PR #1169, opened 2025-07-25
Federator adjustments and enhancements

tpantelis:
See individual commits for details. Highlights:

- Add `IdentifyingLabels` to the create-or-update `Federator` to handle updating resources that don't have a well-known name (eg created via `GenerateName`) and are identified by well-known labels.
- Define a `Federator` implementation with proxy functions to allow users to construct a `Federator` by composition.
- Add a composite `Federator` to encompass multiple `Federators`


pkg/federate/create_or_update_federator.go
```diff
@@ -65,20 +72,75 @@ func (f *createOrUpdateFederator) Distribute(ctx context.Context, obj runtime.Ob
...
 	if f.eventLogName != "" {
 		if result == util.OperationResultCreated {
 			logger.Infof("%s: Created %s \"%s/%s\" ", f.eventLogName, toDistribute.GetKind(), toDistribute.GetNamespace(),
-				toDistribute.GetName())
+				newObj.GetName())
 		} else if result == util.OperationResultUpdated {
 			logger.Infof("%s: Updated %s \"%s/%s\" ", f.eventLogName, toDistribute.GetKind(), toDistribute.GetNamespace(),
-				toDistribute.GetName())
+				newObj.GetName())
 		}
```
skitt: This won’t make a practical difference, but for consistency, it would be better to use `newObj` throughout instead of mixing `toDistribute` and `newObj` (since the created or updated object is `newObj`).

pkg/federate/create_or_update_federator.go
```diff
@@ -65,20 +72,75 @@ func (f *createOrUpdateFederator) Distribute(ctx context.Context, obj runtime.Ob
...
 	}
 
 	return err
 }
+
+func (f *createOrUpdateFederator) Delete(ctx context.Context, obj runtime.Object) error {
+	objMeta := resource.MustToMeta(obj)
+	if objMeta.GetGenerateName() != "" && len(f.identifyingLabels) > 0 {
```
skitt: Can’t we skip this if `obj` already has a name?

skitt (changes requested):

pkg/federate/create_or_update_federator.go
tpantelis: I had it check the name originally but that failed a cleanup test in lighthouse. The usage of this method in the syncer is to pass the object transformed from the object to sync, ie the transformation clears the `Name` and sets the `GenerateName`. The cleanup in lighthouse also calls this method but it didn't clear the `Name` so it tried to delete with the incorrect `Name`. But I've since changed the cleanup to run thru the same transformation so it no longer matters if this checks `Name` and/or `GenerateName`. But, for general usage, it seems better and more intuitive to check if the `Name` is empty.

skitt (approved):

PR #1192, opened 2025-09-12
Signer and CSR implementation

aswinsuryan:
Signer and CSR implementation

### DIFF
--- a/pkg/certificate/signer.go
+++ b/pkg/certificate/signer.go
@@ -19,19 +19,42 @@ limitations under the License.
 package certificate
 
 import (
+	"context"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"math/big"
+	"strconv"
 	"sync"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/federate"
+	"github.com/submariner-io/admiral/pkg/log"
 	"github.com/submariner-io/admiral/pkg/resource"
 	"github.com/submariner-io/admiral/pkg/syncer"
 	"github.com/submariner-io/admiral/pkg/util"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/rest"
+)
+
+const (
+	CASecretName        = "submariner-ca"
+	CertValidity        = 365 * 24 * time.Hour      // 1 year
+	CACertValidity      = 10 * 365 * 24 * time.Hour // 10 years
+	RotateBefore        = 90 * 24 * time.Hour       // 90 days
+	CACheckInterval     = 24 * time.Hour            // Check CA daily
+	CAKeyFileName       = "ca.key"
+	CACertFileName      = "ca.crt"
+	CAVersionAnnotation = "submariner.io/ca-version"
 )
 
 type SignerConfig struct {
@@ -41,7 +64,7 @@ type SignerConfig struct {
 }
 
 type Signer interface {
-	Start(namespace string) error
+	Start(ctx context.Context, namespace string) error
 	Stop(namespace string)
 }
 
@@ -71,11 +94,17 @@ func NewSigner(config SignerConfig) (Signer, error) {
 	return s, nil
 }
 
-func (s *signerImpl) Start(namespace string) error {
+func (s *signerImpl) Start(ctx context.Context, namespace string) error {
 	if _, exists := s.syncerMap.Load(namespace); exists {
 		return nil
 	}
 
+	// Issue/check CA certificate before starting the syncer
+	if err := s.issueCA(ctx, namespace); err != nil {
+		return errors.Wrapf(err, "error issuing CA certificate for namespace %s", namespace)
+	}
+
+	//nolint:contextcheck // NewResourceSyncer doesn't accept context parameter
 	secretSyncer, err := syncer.NewResourceSyncer(&syncer.ResourceSyncerConfig{
 		Name:            "Cert Signer",
 		SourceClient:    s.dynClient,
@@ -88,7 +117,7 @@ func (s *signerImpl) Start(namespace string) error {
 		Transform: func(from runtime.Object, _ int, _ syncer.Operation) (runtime.Object, bool) {
 			secret := from.(*corev1.Secret)
 
-			err := s.signSecret(secret)
+			err := s.signSecret(ctx, secret)
 			if err != nil {
 				logger.Errorf(err, "error signing Secret %q", secret.Name)
 			}
@@ -121,7 +150,12 @@ func (s *signerImpl) Start(namespace string) error {
 		return errors.Wrap(err, "error starting resource syncer")
 	}
 
-	s.syncerMap.Store(namespace, stopCh)
+	if _, loaded := s.syncerMap.LoadOrStore(namespace, stopCh); loaded {
+		return nil // Already started for this namespace
+	}
+
+	// Start periodic CA check for this namespace
+	s.startPeriodicCACheck(ctx, namespace, stopCh)
 
 	return nil
 }
@@ -132,12 +166,298 @@ func (s *signerImpl) Stop(namespace string) {
 	}
 }
 
-// TODO - implement
-//
-//nolint:unparam // (error) is always nil - remove once implemented
-func (s *signerImpl) signSecret(secret *corev1.Secret) error {
-	secret.Data[TLSDataKey] = []byte("tls-" + string(secret.Data[CSRDataKey]))
-	secret.Data[CADataKey] = []byte("ca-" + string(secret.Data[CSRDataKey]))
+func (s *signerImpl) signSecret(ctx context.Context, secret *corev1.Secret) error {
+	csrPEM := secret.Data[CSRDataKey]
+	if len(csrPEM) == 0 {
+		logger.Warning("CSR data is empty for secret \"%s/%s\"", secret.Namespace, secret.Name)
+		return nil
+	}
+
+	// Get CA secret using dynamic client
+	caSecretClient := s.dynClient.Resource(corev1.SchemeGroupVersion.WithResource("secrets")).Namespace(secret.Namespace)
+	caSecretUnstructured, err := caSecretClient.Get(ctx, CASecretName, metav1.GetOptions{})
+	if err != nil {
+		return errors.Wrapf(err, "failed to get CA secret for signing secret %q", secret.Name)
+	}
+
+	caSecret := resource.MustFromUnstructured(caSecretUnstructured, &corev1.Secret{})
+
+	caCertPEM := caSecret.Data[CACertFileName]
+	caKeyPEM := caSecret.Data[CAKeyFileName]
+
+	// Parse CA cert
+	caBlock, _ := pem.Decode(caCertPEM)
+	if caBlock == nil {
+		return errors.Errorf("failed to decode CA secret for signing secret \"%s/%s\"", secret.Namespace, secret.Name)
+	}
+
+	caCert, err := x509.ParseCertificate(caBlock.Bytes)
+	if err != nil {
+		return errors.Wrapf(err, "failed to parse CA certificate for signing secret \"%s/%s\"", secret.Namespace, secret.Name)
+	}
+
+	// Parse CA private key
+	keyBlock, _ := pem.Decode(caKeyPEM)
+	if keyBlock == nil {
+		return errors.Errorf("failed to decode CA private key for signing secret \"%s/%s\"", secret.Namespace, secret.Name)
+	}
+
+	caKey, err := x509.ParsePKCS1PrivateKey(keyBlock.Bytes)
+	if err != nil {
+		return errors.Errorf("failed to parse CA private key for signing secret \"%s/%s\"", secret.Namespace, secret.Name)
+	}
+
+	// Parse CSR
+	csrBlock, _ := pem.Decode(csrPEM)
+	if csrBlock == nil {
+		return errors.Errorf("failed to decode CSR PEM for secret \"%s/%s\"", secret.Namespace, secret.Name)
+	}
+
+	csr, err := x509.ParseCertificateRequest(csrBlock.Bytes)
+	if err != nil {
+		return errors.Errorf("failed to parse CSR PEM for secret \"%s/%s\"", secret.Namespace, secret.Name)
+	}
+
+	if err := csr.CheckSignature(); err != nil {
+		return errors.Wrapf(err, "CSR signature invalid for secret \"%s/%s\"", secret.Namespace, secret.Name)
+	}
+
+	// Create certificate from CSR
+	serialNumber, err := rand.Int(rand.Reader, new(big.Int).Lsh(big.NewInt(1), 128))
+	if err != nil {
+		return errors.Wrapf(err, "failed to generate serial number for secret \"%s/%s\"", secret.Namespace, secret.Name)
+	}
+
+	certTemplate := &x509.Certificate{
+		SerialNumber: serialNumber,
+		Subject:      csr.Subject,
+		NotBefore:    time.Now().Add(-5 * time.Minute),
+		NotAfter:     time.Now().Add(CertValidity),
+		KeyUsage:     x509.KeyUsageDigitalSignature | x509.KeyUsageKeyEncipherment | x509.KeyUsageDataEncipherment,
+		ExtKeyUsage:  []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
+		IPAddresses:  csr.IPAddresses, // Copy IP SANs from CSR
+		DNSNames:     csr.DNSNames,    // Copy DNS SANs from CSR
+	}
+
+	certDER, err := x509.CreateCertificate(rand.Reader, certTemplate, caCert, csr.PublicKey, caKey)
+	if err != nil {
+		return errors.Wrapf(err, "failed to sign CSR for secret \"%s/%s\"", secret.Namespace, secret.Name)
+	}
+
+	signedCertPEM := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: certDER})
+
+	secret.Data[TLSDataKey] = signedCertPEM
+	secret.Data[CADataKey] = caCertPEM
+
+	return nil
+}
+
+// issueCA issues a CA certificate in the specified namespace, creating or rotating it if necessary.
+func (s *signerImpl) issueCA(ctx context.Context, namespace string) error {
+	caSecretClient := s.dynClient.Resource(corev1.SchemeGroupVersion.WithResource("secrets")).Namespace(namespace)
+
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      CASecretName,
+			Namespace: namespace,
+			Annotations: map[string]string{
+				CAVersionAnnotation: "1", // Default version for new CA
+			},
+		},
+	}
+
+	result, _, err := util.CreateOrUpdateWithOptions(ctx, util.CreateOrUpdateOptions[*unstructured.Unstructured]{
+		Client: resource.ForDynamic(caSecretClient),
+		Obj:    resource.MustToUnstructured(secret),
+		MutateOnCreate: func(obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+			secret := resource.MustFromUnstructured(obj, &corev1.Secret{})
+			err := s.signCASecret(secret, secret.Annotations[CAVersionAnnotation])
+			return resource.MustToUnstructured(secret), err
+		},
+		MutateOnUpdate: func(obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+			existing := resource.MustFromUnstructured(obj, &corev1.Secret{})
+
+			shouldReissue, newVersion := s.shouldReissueCA(existing, namespace)
+
+			if !shouldReissue {
+				return obj, nil // No update needed
+			}
+
+			err := s.signCASecret(existing, newVersion)
+			return resource.MustToUnstructured(existing), err
+		},
+	})
+	if err != nil {
+		return errors.Wrapf(err, "failed to create or update CA Secret")
+	}
+
+	if result == util.OperationResultCreated {
+		logger.Infof("Successfully created CA Secret %s in namespace %s", secret.Name, namespace)
+	} else if result == util.OperationResultUpdated {
+		logger.Infof("Successfully rotated CA Secret %s in namespace %s", secret.Name, namespace)
+	}
+
+	// Re-sign all existing CSRs with the new CA (handles both creation and rotation cases)
+	if err := s.resignAllCSRs(ctx, namespace); err != nil {
+		logger.Error(err, "Failed to re-sign existing CSRs with new CA", "namespace", namespace)
+		// Don't return error - CA creation succeeded, CSR re-signing is best effort
+	}
+
+	return nil
+}
+
+// signCASecret generates and signs a CA certificate, storing it in the provided secret.
+func (s *signerImpl) signCASecret(secret *corev1.Secret, version string) error {
+	// Generate new CA certificate
+	privateKey, err := rsa.GenerateKey(rand.Reader, RSABitSize)
+	if err != nil {
+		return errors.Wrapf(err, "failed to generate RSA key")
+	}
+
+	serialNumber, err := rand.Int(rand.Reader, new(big.Int).Lsh(big.NewInt(1), 128))
+	if err != nil {
+		return errors.Wrapf(err, "failed to generate serial number")
+	}
+
+	certTemplate := x509.Certificate{
+		SerialNumber: serialNumber,
+		Subject: pkix.Name{
+			CommonName:   "submariner-ca",
+			Organization: []string{"submariner.io"},
+		},
+		NotBefore:             time.Now().Add(-5 * time.Minute),
+		NotAfter:              time.Now().Add(CACertValidity),
+		KeyUsage:              x509.KeyUsageCertSign | x509.KeyUsageCRLSign,
+		BasicConstraintsValid: true,
+		IsCA:                  true,
+	}
+
+	certDER, err := x509.CreateCertificate(rand.Reader, &certTemplate, &certTemplate, &privateKey.PublicKey, privateKey)
+	if err != nil {
+		return errors.Wrapf(err, "failed to create CA certificate")
+	}
+
+	keyPEM := pem.EncodeToMemory(&pem.Block{Type: "RSA PRIVATE KEY", Bytes: x509.MarshalPKCS1PrivateKey(privateKey)})
+	certPEM := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: certDER})
+
+	if secret.Data == nil {
+		secret.Data = map[string][]byte{}
+	}
+
+	secret.Data[CAKeyFileName] = keyPEM
+	secret.Data[CACertFileName] = certPEM
+
+	// Ensure annotations exist and set the version
+	if secret.Annotations == nil {
+		secret.Annotations = map[string]string{}
+	}
+
+	secret.Annotations[CAVersionAnnotation] = version
+
+	return nil
+}
+
+// shouldReissueCA checks if the CA certificate should be reissued based on expiration time.
+// Returns (shouldReissue, newVersion, error).
+func (s *signerImpl) shouldReissueCA(secret *corev1.Secret, namespace string) (bool, string) {
+	// Get current version from existing secret
+	currentVersion := "1"
+	if existingVersion, exists := secret.Annotations[CAVersionAnnotation]; exists {
+		currentVersion = existingVersion
+	}
+
+	certPEM := secret.Data[CACertFileName]
+	if len(certPEM) == 0 {
+		logger.Warning("CA certificate data is empty, re-issuing for %v", namespace)
+		return true, currentVersion
+	}
+
+	block, _ := pem.Decode(certPEM)
+	if block == nil {
+		logger.Warning("Failed to decode existing CA cert PEM, re-issuingfor %v", namespace)
+		return true, currentVersion
+	}
+
+	cert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		logger.Warning("Failed to parse existing CA cert, re-issuing for %v", namespace)
+		return true, currentVersion
+	}
+
+	timeRemaining := time.Until(cert.NotAfter)
+	logger.V(log.TRACE).Info("Existing CA", "namespace", namespace, "expiresIn", timeRemaining.String(),
+		"notAfter", cert.NotAfter.Format(time.RFC3339), "version", currentVersion)
+
+	if timeRemaining < RotateBefore {
+		newVersion := s.incrementVersion(currentVersion)
+		logger.Infof("CA is expiring soon — rotating it for namespace %s from version %s to %s", namespace, currentVersion, newVersion)
+
+		return true, newVersion
+	}
+
+	logger.V(log.TRACE).Info("CA is still valid — no rotation needed", "namespace", namespace, "version", currentVersion)
+
+	return false, currentVersion
+}
+
+// incrementVersion increments the CA version string, handling both numeric and non-numeric versions.
+func (s *signerImpl) incrementVersion(currentVersion string) string {
+	if version, err := strconv.Atoi(currentVersion); err == nil {
+		return strconv.Itoa(version + 1)
+	}
+
+	// Fallback if current version is not a number
+	logger.Warningf("Current CA version %s is not a number, falling back to version 1", currentVersion)
+
+	return "1"
+}
+
+// startPeriodicCACheck starts a periodic check for CA certificate expiration.
+func (s *signerImpl) startPeriodicCACheck(ctx context.Context, namespace string, stopCh <-chan struct{}) {
+	go wait.Until(func() {
+		logger.V(log.TRACE).Info("Performing periodic CA check", "namespace", namespace)
+
+		// CA exists, check if it needs rotation
+		if err := s.issueCA(ctx, namespace); err != nil {
+			logger.Error(err, "failed periodic CA check", "namespace", namespace)
+		}
+	}, CACheckInterval, stopCh)
+
+	logger.Infof("Started periodic CA check for namespace %s with interval %s", namespace, CACheckInterval)
+}
+
+// resignAllCSRs finds all existing CSR secrets and re-signs them with the new CA.
+func (s *signerImpl) resignAllCSRs(ctx context.Context, namespace string) error {
+	secretClient := s.dynClient.Resource(corev1.SchemeGroupVersion.WithResource("secrets")).Namespace(namespace)
+
+	// List all secrets with the CSR request label
+	secretList, err := secretClient.List(ctx, metav1.ListOptions{
+		LabelSelector: SigningRequestLabelKey,
+	})
+	if err != nil {
+		return errors.Wrapf(err, "failed to list CSR secrets")
+	}
+
+	resignCount := 0
+
+	for _, item := range secretList.Items {
+		err := util.Update(ctx, resource.ForDynamic(secretClient), &item,
+			func(existing *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+				annotations := existing.GetAnnotations()
+				delete(annotations, RequestSignedLabelKey)
+				existing.SetAnnotations(annotations)
+
+				return existing, nil
+			})
+
+		if err != nil {
+			logger.Warning("failed to mark CSR secret \"%s/%s\" for re-signing: %v", err, namespace, item.GetName())
+		} else {
+			logger.Infof("Marked CSR secret \"%s/%s\" for re-signing", namespace, item.GetName())
+		}
+
+		resignCount++
+	}
 
 	return nil
 }
--- a/pkg/certificate/signer_test.go
+++ b/pkg/certificate/signer_test.go
@@ -19,6 +19,14 @@ limitations under the License.
 package certificate_test
 
 import (
+	"context"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"net"
+
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/submariner-io/admiral/pkg/certificate"
@@ -41,9 +49,11 @@ var _ = Describe("Signer", func() {
 	var (
 		dynClient *dynamicfake.FakeDynamicClient
 		signer    certificate.Signer
+		ctx       context.Context
 	)
 
 	BeforeEach(func() {
+		ctx = context.Background()
 		dynClient = dynamicfake.NewSimpleDynamicClient(scheme.Scheme)
 
 		resource.NewDynamicClient = func(_ *rest.Config) (dynamic.Interface, error) {
@@ -65,7 +75,7 @@ var _ = Describe("Signer", func() {
 		})
 		Expect(err).NotTo(HaveOccurred())
 
-		Expect(signer.Start(namespace)).To(Succeed())
+		Expect(signer.Start(ctx, namespace)).To(Succeed())
 	})
 
 	AfterEach(func() {
@@ -75,7 +85,7 @@ var _ = Describe("Signer", func() {
 	When("a CSR Secret is created and updated", func() {
 		It("should sign it", func() {
 			// Starting again should be a no-op.
-			Expect(signer.Start(namespace)).To(Succeed())
+			Expect(signer.Start(ctx, namespace)).To(Succeed())
 
 			client := secretClient(dynClient, namespace)
 
@@ -96,7 +106,7 @@ var _ = Describe("Signer", func() {
 
 			By("Updating the signed Secret")
 
-			secret.Data[certificate.CSRDataKey] = []byte("csr-data-updated")
+			secret.Data[certificate.CSRDataKey] = generateTestCSR()
 			delete(secret.Annotations, certificate.RequestSignedLabelKey)
 
 			test.UpdateResource(client, secret)
@@ -136,7 +146,7 @@ var _ = Describe("Signer", func() {
 			client1 := secretClient(dynClient, namespace)
 			client2 := secretClient(dynClient, namespace2)
 
-			Expect(signer.Start(namespace2)).To(Succeed())
+			Expect(signer.Start(ctx, namespace2)).To(Succeed())
 
 			secret := newCSR()
 
@@ -178,6 +188,9 @@ var _ = Describe("Signer", func() {
 })
 
 func newCSR() *corev1.Secret {
+	// Generate valid CSR data
+	csrPEM := generateTestCSR()
+
 	return &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: "ipsec",
@@ -185,6 +198,31 @@ func newCSR() *corev1.Secret {
 				certificate.SigningRequestLabelKey: "east",
 			},
 		},
-		Data: map[string][]byte{certificate.CSRDataKey: []byte("csr-data")},
+		Data: map[string][]byte{certificate.CSRDataKey: csrPEM},
 	}
+}
+
+func generateTestCSR() []byte {
+	// Generate a test private key
+	privateKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	Expect(err).NotTo(HaveOccurred())
+
+	// Create CSR template
+	template := x509.CertificateRequest{
+		Subject: pkix.Name{
+			CommonName:   "test-cert",
+			Organization: []string{"submariner.io"},
+		},
+		SignatureAlgorithm: x509.SHA256WithRSA,
+		IPAddresses:        []net.IP{net.ParseIP("192.168.1.1")},
+	}
+
+	// Create CSR
+	csrDER, err := x509.CreateCertificateRequest(rand.Reader, &template, privateKey)
+	Expect(err).NotTo(HaveOccurred())
+
+	// Encode to PEM
+	csrPEM := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE REQUEST", Bytes: csrDER})
+
+	return csrPEM
 }
--- a/pkg/certificate/signing_requestor.go
+++ b/pkg/certificate/signing_requestor.go
@@ -21,14 +21,22 @@ package certificate
 import (
 	"bytes"
 	"context"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
 	goerrors "errors"
 	"fmt"
+	"net"
 	"sync"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/federate"
 	"github.com/submariner-io/admiral/pkg/log"
 	"github.com/submariner-io/admiral/pkg/resource"
+	"github.com/submariner-io/admiral/pkg/slices"
 	"github.com/submariner-io/admiral/pkg/syncer"
 	"github.com/submariner-io/admiral/pkg/syncer/broker"
 	"github.com/submariner-io/admiral/pkg/util"
@@ -39,6 +47,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	k8slabels "k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/dynamic"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
@@ -47,14 +56,26 @@ const (
 	SigningRequestLabelKey = "submariner.io/csr-request"
 	RequestSignedLabelKey  = "submariner.io/csr-request-signed"
 	PrivateKeyDataKey      = "tls.key"
+	RSABitSize             = 2048
 	CSRDataKey             = "csr.pem"
 	TLSDataKey             = "tls.crt"
 	CADataKey              = "ca.crt"
+
+	// Certificate renewal constants.
+	CertRenewBefore   = 30 * 24 * time.Hour // Renew 30 days before expiration
+	CertCheckInterval = 12 * time.Hour      // Check certificate expiration every 12 hours
 )
 
 type OnSignedFn func(secretData map[string][]byte) error
 
 type KeyGeneratorFn func(ips []string) ([]byte, []byte, error)
+
+type certInfo struct {
+	name      string
+	ips       []string
+	onSigned  OnSignedFn
+	expiresAt time.Time
+}
 
 type SigningRequestor interface {
 	Issue(ctx context.Context, name string, ips []string, onSigned OnSignedFn) error
@@ -70,8 +91,10 @@ type signingRequestorImpl struct {
 	localClusterID     string
 	localSecretClient  dynamic.ResourceInterface
 	brokerSecretClient dynamic.ResourceInterface
-	onSignedMap        sync.Map
 	keyGenerator       KeyGeneratorFn
+
+	// Certificate management fields
+	issuedCerts sync.Map // map[secretName]certInfo
 }
 
 var logger = log.Logger{Logger: logf.Log.WithName("Certificate")}
@@ -179,6 +202,9 @@ func StartSigningRequestor(syncerConfig broker.SyncerConfig, stopCh <-chan struc
 		return nil, errors.Wrap(err, "error starting secret watcher")
 	}
 
+	// Start certificate renewal monitoring
+	sr.startCertificateRenewalMonitoring(stopCh)
+
 	return sr, nil
 }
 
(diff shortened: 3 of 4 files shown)


pkg/certificate/signing_requestor.go
```diff
@@ -0,0 +1,487 @@
...
+	localClusterID     string
+	localSecretClient  dynamic.ResourceInterface
+	brokerSecretClient dynamic.ResourceInterface
+	onSignedMap        sync.Map
+	keyGenerator       KeyGeneratorFn
+
+	// Certificate renewal fields
+	renewalTicker *time.Ticker
+	renewalStopCh chan struct{}
+	issuedCerts   sync.Map // map[secretName]certRenewalInfo
```
tpantelis: Since this also has the `OnSignedFn` function, it can replace the `onSignedMap`. We can also rename `certRenewalInfo` to just `certInfo`.

pkg/certificate/signing_requestor.go
```diff
@@ -0,0 +1,487 @@
...
+		return nil
+	}
+
+	return errors.Wrapf(err, "error deleting Secret %q", name)
+}
+
+// startCertificateRenewalMonitoring starts periodic monitoring of certificate expiration
+func (s *signingRequestorImpl) startCertificateRenewalMonitoring(stopCh <-chan struct{}) {
+	s.renewalTicker = time.NewTicker(CertCheckInterval)
+	s.renewalStopCh = make(chan struct{})
```
tpantelis: We have the user-provided `stopCh` so I don't think we need the `renewalStopCh`.

pkg/certificate/signing_requestor.go
```diff
@@ -0,0 +1,487 @@
...
+			select {
+			case <-s.renewalTicker.C:
+				s.checkCertificateRenewal()
+			case <-s.renewalStopCh:
+				return
+			case <-stopCh:
+				return
+			}
+		}
+	}()
```
tpantelis: You can use `wait.Until` instead:

```suggestion
	wait.Until(s.checkCertificateRenewal, CertCheckInterval, stopCh)
```

pkg/certificate/signing_requestor.go
```diff
@@ -0,0 +1,487 @@
...
+	s.onSignedMap.Delete(secretName)
+	s.issuedCerts.Delete(secretName)
+
+	return goerrors.Join(deleteIfPresent(ctx, s.localSecretClient, secretName),
+		deleteIfPresent(ctx, s.brokerSecretClient, secretName))
+}
+
+func (s *signingRequestorImpl) Uninstall(ctx context.Context) error {
+	// Stop certificate renewal monitoring
+	s.stopCertificateRenewalMonitoring()
```
tpantelis: We don't really need to do this since the monitoring will be stopped when the user closes the `stopCh`. `Uninstall` is intended to do cleanup.

pkg/certificate/signing_requestor.go
```diff
@@ -0,0 +1,487 @@
...
+// updateCertificateExpiration extracts and stores the certificate expiration time
+func (s *signingRequestorImpl) updateCertificateExpiration(secret *corev1.Secret) {
+	certPEM := secret.Data[TLSDataKey]
+	if len(certPEM) == 0 {
+		return
+	}
+
+	block, _ := pem.Decode(certPEM)
+	if block == nil {
+		logger.Warningf("Failed to decode certificate PEM for secret %q", secret.Name)
```
tpantelis: Shouldn't this be an error?

```suggestion
		logger.Errorf("Failed to find certificate PEM in %q for secret %q", TLSDataKey, secret.Name)
```

pkg/certificate/signing_requestor.go
```diff
@@ -0,0 +1,487 @@
...
+		return
+	}
+
+	// Update the expiration time in our tracking map
+	if v, ok := s.issuedCerts.Load(secret.Name); ok {
+		renewalInfo := v.(certRenewalInfo)
+		renewalInfo.expiresAt = cert.NotAfter
+		s.issuedCerts.Store(secret.Name, renewalInfo)
+
+		logger.V(1).Infof("Updated certificate expiration for %q: expires at %s",
```
tpantelis: I think level 1 is the default.

```suggestion
		logger.Infof("Updated certificate expiration for secret %q: expires at %s",
```

pkg/certificate/signing_requestor.go
```diff
@@ -0,0 +1,487 @@
...
+			logger.Infof("Certificate %q expires in %s, renewing", secretName, timeUntilExpiry.String())
+
+			// Renew the certificate by re-issuing it
+			if err := s.Issue(context.TODO(), renewalInfo.name, renewalInfo.ips, renewalInfo.onSigned); err != nil {
+				logger.Errorf(err, "Failed to renew certificate %q", secretName)
+			} else {
+				logger.Infof("Successfully initiated renewal for certificate %q", secretName)
+			}
+		} else {
+			logger.V(2).Infof("Certificate %q expires in %s, no renewal needed", secretName, timeUntilExpiry.String())
```
tpantelis: We should use our log constants. I think this should be TRACE.

```suggestion
			logger.V(log.TRACE).Infof("Certificate %q expires in %s, no renewal needed", secretName, timeUntilExpiry.String())
```

tpantelis (changes requested):

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+	secret.Data[TLSDataKey] = signedCertPEM
+	secret.Data[CADataKey] = caCertPEM
+
+	return nil
+}
+
+// IssueCA issues a CA certificate in the specified namespace, creating or rotating it if necessary.
+func (s *signerImpl) IssueCA(ctx context.Context, namespace string) error {
+	// Store the namespace for use in signing operations
+	s.namespace = namespace
```
tpantelis: Since we have signers per namespace, we don't want a shared namespace.

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+}
+
+func (s *signerImpl) signSecret(secret *corev1.Secret) error {
+	csrPEM := secret.Data[CSRDataKey]
+	if len(csrPEM) == 0 {
+		return fmt.Errorf("CSR data is empty")
+	}
+
+	// Get CA secret using dynamic client
+	caSecretClient := s.dynClient.Resource(corev1.SchemeGroupVersion.WithResource("secrets")).Namespace(s.namespace)
```
tpantelis: Since we have signers per namespace, we want to use the `Namespace` from the `secret`.

```suggestion
	caSecretClient := s.dynClient.Resource(corev1.SchemeGroupVersion.WithResource("secrets")).Namespace(secret.Namespace)
```

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+	caSecretUnstructured, err := caSecretClient.Get(context.TODO(), CASecretName, metav1.GetOptions{})
+	if err != nil {
+		return errors.Wrapf(err, "failed to get CA secret for signing secret %q", secret.Name)
+	}
+
+	caSecret := &corev1.Secret{}
+	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(caSecretUnstructured.Object, caSecret); err != nil {
+		return errors.Wrapf(err, "failed to convert CA secret for signing secret %q", secret.Name)
+	}
```
tpantelis:
```suggestion
	caSecret := := resource.MustFromUnstructured(caSecretUnstructured, &corev1.Secret{})
```

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+
+	// Stop periodic CA check when stopping the signer
+	s.stopPeriodicCACheck()
+}
+
+func (s *signerImpl) signSecret(secret *corev1.Secret) error {
+	csrPEM := secret.Data[CSRDataKey]
+	if len(csrPEM) == 0 {
+		return fmt.Errorf("CSR data is empty")
```
tpantelis:
```suggestion
		return errors.New("CSR data is empty for secret \"%s/%s\"", secret.Namespace, secret.Name)
```

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+	}
+
+	// Get CA secret using dynamic client
+	caSecretClient := s.dynClient.Resource(corev1.SchemeGroupVersion.WithResource("secrets")).Namespace(s.namespace)
+	caSecretUnstructured, err := caSecretClient.Get(context.TODO(), CASecretName, metav1.GetOptions{})
+	if err != nil {
+		return errors.Wrapf(err, "failed to get CA secret for signing secret %q", secret.Name)
```
tpantelis, line 180:
```suggestion
		return errors.Wrapf(err, "failed to get CA secret for signing secret \"%s/%s\"", secret.Namespace, secret.Name)
```

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+		return errors.Wrapf(err, "failed to convert CA secret for signing secret %q", secret.Name)
+	}
+
+	caCertPEM := caSecret.Data[CACertFileName]
+	caKeyPEM := caSecret.Data[CAKeyFileName]
+
+	// Parse CA cert
+	caBlock, _ := pem.Decode(caCertPEM)
+	if caBlock == nil {
+		return errors.Wrapf(fmt.Errorf("failed to decode CA certificate"), "failed to sign CSR for secret %q", secret.Name)
```
tpantelis:
```suggestion
		return errors.Errorf("failed to decode CA secret for signing secret \"%s/%s\"", secret.Namespace, secret.Name)
```

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+	}
+
+	caCert, err := x509.ParseCertificate(caBlock.Bytes)
+	if err != nil {
+		return errors.Wrapf(err, "failed to parse CA certificate for secret %q", secret.Name)
```
tpantelis:
```suggestion
		return errors.Wrapf(err, "failed to parse CA certificate for signing secret \"%s/%s\"", secret.Namespace, secret.Name)
```

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+	}
+
+	// Parse CA private key
+	keyBlock, _ := pem.Decode(caKeyPEM)
+	if keyBlock == nil {
+		return errors.Wrapf(fmt.Errorf("failed to decode CA private key"), "failed to sign CSR for secret %q", secret.Name)
```
tpantelis:
```suggestion
        return errors.Errorf("failed to decode CA private key for signing secret \"%s/%s\"", secret.Namespace, secret.Name)
```

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+	}
+
+	caKey, err := x509.ParsePKCS1PrivateKey(keyBlock.Bytes)
+	if err != nil {
+		return errors.Wrapf(err, "failed to parse CA private key for secret %q", secret.Name)
```
tpantelis:
```suggestion
		return errors.Errorf("failed to parse CA private key for signing secret \"%s/%s\"", secret.Namespace, secret.Name)
```

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+	}
+
+	// Parse CSR
+	csrBlock, _ := pem.Decode(csrPEM)
+	if csrBlock == nil {
+		return errors.Wrapf(fmt.Errorf("failed to decode CSR PEM"), "failed to sign CSR for secret %q", secret.Name)
```
tpantelis:
```suggestion
		return errors.Errorf("failed to decode CSR PEM for secret \"%s/%s\"", secret.Namespace, secret.Name)
```

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+	}
+
+	csr, err := x509.ParseCertificateRequest(csrBlock.Bytes)
+	if err != nil {
+		return errors.Wrapf(err, "failed to parse CSR for secret %q", secret.Name)
```
tpantelis:
```suggestion
		return return errors.Errorf("failed to parse CSR PEM for secret \"%s/%s\"", secret.Namespace, secret.Name)
```

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+	}
+
+	if err := csr.CheckSignature(); err != nil {
+		return errors.Wrapf(err, "CSR signature invalid for secret %q", secret.Name)
```
tpantelis:
```suggestion
		return errors.Wrapf(err, "CSR signature invalid for secret \"%s/%s\"", secret.Namespace, secret.Name)
```

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+	}
+
+	// Create certificate from CSR
+	serialNumber, err := rand.Int(rand.Reader, new(big.Int).Lsh(big.NewInt(1), 128))
+	if err != nil {
+		return errors.Wrapf(err, "failed to generate serial number for secret %q", secret.Name)
```
tpantelis:
```suggestion
		return errors.Wrapf(err, "failed to generate serial number for secret \"%s/%s\"", secret.Namespace, secret.Name)
```

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+		NotAfter:     time.Now().Add(CertValidity),
+		KeyUsage:     x509.KeyUsageDigitalSignature | x509.KeyUsageKeyEncipherment | x509.KeyUsageDataEncipherment,
+		ExtKeyUsage:  []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
+		IPAddresses:  csr.IPAddresses, // Copy IP SANs from CSR
+		DNSNames:     csr.DNSNames,    // Copy DNS SANs from CSR
+	}
+
+	certDER, err := x509.CreateCertificate(rand.Reader, certTemplate, caCert, csr.PublicKey, caKey)
+	if err != nil {
+		return errors.Wrapf(err, "failed to sign CSR for secret %q", secret.Name)
```
tpantelis:
```suggestion
		return errors.Wrapf(err, "failed to sign CSR for secret \"%s/%s\"", secret.Namespace, secret.Name)
```

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+		}
+
+		// Secret doesn't exist, create new one
+		logger.Info("CA Secret not found — generating new CA", "namespace", namespace)
+		secret = &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      CASecretName,
+				Namespace: namespace,
+				Labels: map[string]string{
+					"submariner.io/component": "ca",
```
tpantelis: What is this for?

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+		if err := s.resignAllCSRs(ctx, namespace); err != nil {
+			logger.Error(err, "Failed to re-sign existing CSRs with new CA", "namespace", namespace)
+			// Don't return error - CA creation succeeded, CSR re-signing is best effort
+		}
+	}
+
+	// Generate new CA certificate
+	privateKey, err := rsa.GenerateKey(rand.Reader, RSABitSize)
+	if err != nil {
+		return fmt.Errorf("failed to generate RSA key: %w", err)
```
tpantelis:
```suggestion
		return errors.Wrapf(err, "failed to generate RSA key")
```

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+	}
+
+	serialNumber, err := rand.Int(rand.Reader, new(big.Int).Lsh(big.NewInt(1), 128))
+	if err != nil {
+		return fmt.Errorf("failed to generate serial number: %w", err)
```
tpantelis:
```suggestion
		return errors.Wrapf(err, "failed to generate serial number")
```

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+		NotBefore:             time.Now().Add(-5 * time.Minute),
+		NotAfter:              time.Now().Add(CACertValidity),
+		KeyUsage:              x509.KeyUsageCertSign | x509.KeyUsageCRLSign,
+		BasicConstraintsValid: true,
+		IsCA:                  true,
+	}
+
+	certDER, err := x509.CreateCertificate(rand.Reader, &certTemplate, &certTemplate, &privateKey.PublicKey, privateKey)
+	if err != nil {
+		return fmt.Errorf("failed to create CA certificate: %w", err)
```
tpantelis:
```suggestion
		return errors.Wrapf(err,, "failed to create CA certificate")
```

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+		logger.Info("CA Secret not found — generating new CA", "namespace", namespace)
+		secret = &corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      CASecretName,
+				Namespace: namespace,
+				Labels: map[string]string{
+					"submariner.io/component": "ca",
+				},
+				Annotations: map[string]string{
+					"submariner.io/ca-version": currentVersion,
```
tpantelis: Let's make a constant for "submariner.io/ca-version".

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+	}
+	secret.Annotations["submariner.io/ca-version"] = currentVersion
+
+	// Create or update the secret
+	secretUnstructured := resource.MustToUnstructured(secret)
+	if secret.CreationTimestamp.IsZero() {
+		_, err = caSecretClient.Create(ctx, secretUnstructured, metav1.CreateOptions{})
+	} else {
+		_, err = caSecretClient.Update(ctx, secretUnstructured, metav1.UpdateOptions{})
+	}
```
tpantelis: We can use `util.CreateOrUpdateWithOptions` here:

```
secret = &corev1.Secret{
		ObjectMeta: metav1.ObjectMeta{
			Name:      CASecretName,
			Namespace: namespace,
			Labels: map[string]string{
				"submariner.io/component": "ca",
			},
			Annotations: map[string]string{
				"submariner.io/ca-version": currentVersion,
			},
		},
	}

result, _, err := util.CreateOrUpdateWithOptions(ctx, util.CreateOrUpdateOptions[*unstructured.Unstructured]{
		Client: resource.ForDynamic(caSecretClient),
		Obj:    resource.MustToUnstructured(secret),
		MutateOnCreate: func(obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
			secret := resource.MustFromUnstructured(obj, &corev1.Secret{})
			err := s.signCASecret(secret, "1")

			return resource.MustToUnstructured(secret), err
		},
		MutateOnUpdate: func(obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
			existing := resource.MustFromUnstructured(obj, &corev1.Secret{})

			shouldReissue, newVersion, err := s.shouldReissueCA(existing, namespace)
			...

			err := s.signCASecret(secret, newVersion)

			return resource.MustToUnstructured(existing), err
		},
	})

	if result == util.OperationResultCreated {
		logger.Infof("Successfully created CA Secret \"%s/%s\"", namespace, secret.Name)
	} else if result == util.OperationResultUpdated {
		logger.Infof("Successfully rotated CA Secret \"%s/%s\"", namespace, secret.Name)
	}
```

`s.signCASecret` would contain L314-356

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+func (s *signerImpl) shouldReissueCA(secret *corev1.Secret, namespace string) (bool, string, error) {
+	// Get current version from existing secret
+	currentVersion := "1"
+	if existingVersion, exists := secret.Annotations["submariner.io/ca-version"]; exists {
+		currentVersion = existingVersion
+	}
+
+	certPEM := secret.Data[CACertFileName]
+	if len(certPEM) == 0 {
+		logger.Info("CA certificate data is empty, reissuing", "namespace", namespace)
```
tpantelis:
```suggestion
		logger.Info("CA certificate data is empty, re-issuing", "namespace", namespace)
```

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+		return true, s.incrementVersion(currentVersion), nil
+	}
+
+	block, _ := pem.Decode(certPEM)
+	if block == nil {
+		logger.Info("Failed to decode existing CA cert PEM, reissuing", "namespace", namespace)
```
tpantelis: Should this be logged as an error?

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+		return true, s.incrementVersion(currentVersion), nil
+	}
+
+	cert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		logger.Info("Failed to parse existing CA cert, reissuing", "namespace", namespace, "error", err)
```
tpantelis: Same here.

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+		return true, s.incrementVersion(currentVersion), nil
+	}
+
+	timeRemaining := time.Until(cert.NotAfter)
+	logger.Info("Existing CA", "namespace", namespace, "expiresIn", timeRemaining.String(), "notAfter", cert.NotAfter.Format(time.RFC3339), "version", currentVersion)
```
tpantelis: Probably TRACE here.

```suggestion
	logger.V(log.TRACE).Info("Existing CA", "namespace", namespace, "expiresIn", timeRemaining.String(), "notAfter", cert.NotAfter.Format(time.RFC3339), "version", currentVersion)
```

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+
+	if timeRemaining < RotateBefore {
+		logger.Info("CA is expiring soon — rotating it", "namespace", namespace, "currentVersion", currentVersion)
+		newVersion := s.incrementVersion(currentVersion)
+		logger.Info("Incrementing CA version for rotation", "namespace", namespace, "newVersion", newVersion)
+		return true, newVersion, nil
+	}
+
+	logger.Info("CA is still valid — no rotation needed", "namespace", namespace, "version", currentVersion)
```
tpantelis:
```suggestion
	logger.V(log.TRACE).Info("CA is still valid — no rotation needed", "namespace", namespace, "version", currentVersion)
```

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+	stopCh := make(chan struct{})
+
+	if err := secretSyncer.Start(stopCh); err != nil {
+		return errors.Wrap(err, "error starting resource syncer")
+	}
+
+	s.syncerMap.Store(namespace, stopCh)
+
+	// Start periodic CA check if not already running
+	s.startPeriodicCACheck(namespace)
```
tpantelis: We can ensure the CA check is only started once by:

```suggestion
	if _, loaded := s.syncerMap.LoadOrStore(namespace, stopCh); loaded {
		s.startPeriodicCACheck(namespace)
	}
```

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+				if err := s.IssueCA(context.TODO(), namespace); err != nil {
+					logger.Error(err, "Failed periodic CA check", "namespace", namespace)
+				}
+			case <-s.caStopCh:
+				return
+			}
+		}
+	}()
+
+	logger.Info("Started periodic CA check", "interval", CACheckInterval, "namespace", namespace)
```
tpantelis: We can use `wait.Until` (pass in `stopCh`):

```
wait.Until(func() {
	// do periodic check
}, CACheckInterval, stopCh)
```

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+		s.caTicker = nil
+	}
+
+	if s.caStopCh != nil {
+		close(s.caStopCh)
+		s.caStopCh = nil
+	}
+
+	logger.Info("Stopped periodic CA check", "namespace", s.namespace)
+}
```
tpantelis: This isn't needed.

```suggestion
```

pkg/certificate/signer.go
```diff
@@ -0,0 +1,528 @@
...
+			secret.Annotations = map[string]string{}
+		}
+		delete(secret.Annotations, RequestSignedLabelKey)
+
+		// Update the secret to trigger re-signing
+		secretUnstructured := resource.MustToUnstructured(secret)
+		if _, err := secretClient.Update(ctx, secretUnstructured, metav1.UpdateOptions{}); err != nil {
+			logger.Error(err, "Failed to update CSR secret for re-signing", "secret", secret.Name)
+			continue
+		}
```
tpantelis: You can use `util.Update`:

```suggestion
		util.Update(ctx, resource.ForDynamic(secretClient), &item,
			func(existing *unstructured.Unstructured) (*unstructured.Unstructured, error) {
				annotations := existing.GetAnnotations()
				delete(annotations, RequestSignedLabelKey)
				existing.SetAnnotations(annotations)

				return existing, nil
			})
```

tpantelis (changes requested):

pkg/certificate/signer.go
```diff
@@ -121,7 +150,22 @@ func (s *signerImpl) Start(namespace string) error {
 		return errors.Wrap(err, "error starting resource syncer")
 	}
 
-	s.syncerMap.Store(namespace, stopCh)
+	if _, loaded := s.syncerMap.LoadOrStore(namespace, stopCh); loaded {
+		return nil // Already started for this namespace
+	}
+
+	// Issue/check CA certificate when starting
+	if err := s.issueCA(context.TODO(), namespace); err != nil {
```
tpantelis: We should pass a `ctx` into this function.

pkg/certificate/signer.go
```diff
@@ -121,7 +150,22 @@ func (s *signerImpl) Start(namespace string) error {
...
+		return errors.Wrapf(err, "error issuing CA certificate for namespace %s", namespace)
+	}
+
+	// Start periodic CA check if not already running
+	s.caCheckMutex.Lock()
+	if !s.caCheckStarted {
+		s.startPeriodicCACheck(namespace)
+		s.caCheckStarted = true
+	}
+	s.caCheckMutex.Unlock()
```
tpantelis: Here it seems the intention is to have a single CA check across all namespaces but `startPeriodicCACheck` takes the namespace so will only operate on that namespace.   It's simpler to start a CA check per namespace using the `stopCh` created above that's stored in the `syncerMap`:

`s.startPeriodicCACheck(namespace, stopCh)`

pkg/certificate/signer.go
```diff
@@ -130,14 +174,339 @@ func (s *signerImpl) Stop(namespace string) {
...
+	})
+
+	if !hasActiveNamespaces && s.caCheckStarted {
+		if s.caStopCh != nil {
+			close(s.caStopCh)
+			s.caStopCh = nil
+		}
+
+		s.caCheckStarted = false
+	}
```
tpantelis: Per comments above, closing the `chan` on L175 will also stop the CA check.

```suggestion
```

pkg/certificate/signer.go
```diff
@@ -130,14 +174,339 @@ func (s *signerImpl) Stop(namespace string) {
...
 
-// TODO - implement
-//
-//nolint:unparam // (error) is always nil - remove once implemented
 func (s *signerImpl) signSecret(secret *corev1.Secret) error {
-	secret.Data[TLSDataKey] = []byte("tls-" + string(secret.Data[CSRDataKey]))
-	secret.Data[CADataKey] = []byte("ca-" + string(secret.Data[CSRDataKey]))
+	csrPEM := secret.Data[CSRDataKey]
+	if len(csrPEM) == 0 {
+		return errors.Errorf("CSR data is empty for secret \"%s/%s\"", secret.Namespace, secret.Name)
```
tpantelis: This really shouldn't happen  but returning an error will cause it to keep retrying. I think we should just log a warning.

pkg/certificate/signer.go
```diff
@@ -130,14 +174,339 @@ func (s *signerImpl) Stop(namespace string) {
...
+			secret := resource.MustFromUnstructured(obj, &corev1.Secret{})
+			err := s.signCASecret(secret, "1")
+			return resource.MustToUnstructured(secret), err
+		},
+		MutateOnUpdate: func(obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+			existing := resource.MustFromUnstructured(obj, &corev1.Secret{})
+
+			shouldReissue, newVersion, err := s.shouldReissueCA(existing, namespace)
+			if err != nil {
+				return nil, fmt.Errorf("failed to check CA reissue status: %w", err)
```
tpantelis: Use `errors.Wrapf`.

pkg/certificate/signer.go
```diff
@@ -130,14 +174,339 @@ func (s *signerImpl) Stop(namespace string) {
...
+func (s *signerImpl) shouldReissueCA(secret *corev1.Secret, namespace string) (bool, string, error) {
+	// Get current version from existing secret
+	currentVersion := "1"
+	if existingVersion, exists := secret.Annotations[CAVersionAnnotation]; exists {
+		currentVersion = existingVersion
+	}
+
+	certPEM := secret.Data[CACertFileName]
+	if len(certPEM) == 0 {
+		return true, currentVersion, errors.Errorf("CA certificate data is empty, re-issuing for %v", namespace)
```
tpantelis: This really shouldn't happen but returning an error will not cause it to be re-issued. We should just log a warning or error. Same for L412 and L417. That likely means there's something corrupted. In fact, we don't need to return error at all from this function, just log warnings/errors and return true so the caller re-issues.

pkg/certificate/signer.go
```diff
@@ -130,14 +174,339 @@ func (s *signerImpl) Stop(namespace string) {
...
+	if err != nil {
+		return true, currentVersion, errors.Wrapf(err, "Failed to parse existing CA cert, re-issuing for %v", namespace)
+	}
+
+	timeRemaining := time.Until(cert.NotAfter)
+	logger.V(log.TRACE).Info("Existing CA", "namespace", namespace, "expiresIn", timeRemaining.String(),
+		"notAfter", cert.NotAfter.Format(time.RFC3339), "version", currentVersion)
+
+	if timeRemaining < RotateBefore {
+		logger.Info("CA is expiring soon — rotating it", "namespace", namespace, "currentVersion", currentVersion)
```
tpantelis: We can combine L427 with this.

```suggestion
		logger.Info("CA is expiring soon — rotating it", "namespace", namespace, "currentVersion", currentVersion, "newVersion", newVersion)
```

pkg/certificate/signer.go
```diff
@@ -130,14 +174,339 @@ func (s *signerImpl) Stop(namespace string) {
...
+	// Fallback if current version is not a number
+	return "2"
+}
+
+// startPeriodicCACheck starts a periodic check for CA certificate expiration.
+func (s *signerImpl) startPeriodicCACheck(namespace string) {
+	s.caStopCh = make(chan struct{})
+
+	go wait.Until(func() {
+		logger.V(1).Info("Performing periodic CA check", "namespace", namespace)
```
tpantelis:
```suggestion
		logger.V(log.TRACE).Info("Performing periodic CA check", "namespace", namespace)
```

pkg/certificate/signer.go
```diff
@@ -130,14 +174,339 @@ func (s *signerImpl) Stop(namespace string) {
...
+// resignAllCSRs finds all existing CSR secrets and re-signs them with the new CA.
+func (s *signerImpl) resignAllCSRs(ctx context.Context, namespace string) error {
+	secretClient := s.dynClient.Resource(corev1.SchemeGroupVersion.WithResource("secrets")).Namespace(namespace)
+
+	// List all secrets with the CSR request label
+	secretList, err := secretClient.List(ctx, metav1.ListOptions{
+		LabelSelector: SigningRequestLabelKey,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to list CSR secrets: %w", err)
```
tpantelis: Use `errors.Wrapf`.

pkg/certificate/signer.go
```diff
@@ -130,14 +174,339 @@ func (s *signerImpl) Stop(namespace string) {
...
+				annotations := existing.GetAnnotations()
+				delete(annotations, RequestSignedLabelKey)
+				existing.SetAnnotations(annotations)
+
+				return existing, nil
+			})
+
+		resignCount++
+
+		logger.Info("Marked CSR secret for re-signing", "secret", item.GetName(), "namespace", namespace)
```
tpantelis:
```suggestion
		logger.Infof("Marked CSR secret \"%s/%s\" for re-signing", namespace, item.GetName())
```
PS: you're using a mix of logging styles, ie either in-line formatted or with separate key/values. Perhaps we should use one for consistency (I'm partial to in-line formatted).

pkg/certificate/signer.go
```diff
@@ -130,14 +174,339 @@ func (s *signerImpl) Stop(namespace string) {
...
+	}
+
+	if resignCount > 0 {
+		logger.Info("Marked CSR secrets for re-signing with new CA", "count", resignCount, "namespace", namespace)
+	} else {
+		logger.Info("No existing CSR secrets found to re-sign", "namespace", namespace)
+	}
```
tpantelis: I don't think we need this since we logged each CSR above.

pkg/certificate/signer.go
tpantelis: Also it would be ideal to do this before starting the syncer.

pkg/certificate/signer.go
```diff
@@ -130,14 +174,339 @@ func (s *signerImpl) Stop(namespace string) {
...
+
+			if !shouldReissue {
+				return obj, nil // No update needed
+			}
+
+			// Re-sign all existing CSRs with the new CA (only called when CA is actually rotated)
+			if err := s.resignAllCSRs(ctx, namespace); err != nil {
+				logger.Error(err, "Failed to re-sign existing CSRs with new CA", "namespace", namespace)
+				// Don't return error - CA creation succeeded, CSR re-signing is best effort
+			}
```
tpantelis: We should ensure the CA secret is re-signed and updated before calling `resignAllCSRs` to avoid potential timing issue. So it would be safest to call `resignAllCSRs` after the CA secret is actually updated, ie after L339 below.

pkg/certificate/signer.go
```diff
@@ -130,14 +174,339 @@ func (s *signerImpl) Stop(namespace string) {
...
+	return false, currentVersion, nil
+}
+
+// incrementVersion increments the CA version string, handling both numeric and non-numeric versions.
+func (s *signerImpl) incrementVersion(currentVersion string) string {
+	if version, err := strconv.Atoi(currentVersion); err == nil {
+		return strconv.Itoa(version + 1)
+	}
+	// Fallback if current version is not a number
+	return "2"
```
tpantelis: We should log an error here. Also shouldn't we fallback to the default version "1"?

pkg/certificate/signer.go
```diff
@@ -130,14 +174,339 @@ func (s *signerImpl) Stop(namespace string) {
...
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				logger.V(1).Info("CA secret not found during periodic check, skipping", "namespace", namespace)
+				return
+			}
+
+			logger.Error(err, "Failed to check CA secret during periodic check", "namespace", namespace)
+
+			return
+		}
```
tpantelis: We don't need this as `issueCA` will retrieve before create/update.

```suggestion
```

pkg/certificate/signer.go
```diff
@@ -130,14 +174,339 @@ func (s *signerImpl) Stop(namespace string) {
...
+		LabelSelector: SigningRequestLabelKey,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to list CSR secrets: %w", err)
+	}
+
+	resignCount := 0
+
+	for _, item := range secretList.Items {
+		_ = util.Update(ctx, resource.ForDynamic(secretClient), &item,
```
tpantelis: We should log the returned error.

pkg/certificate/signing_requestor.go
```diff
@@ -286,7 +352,12 @@ func (s *signingRequestorImpl) onLocalSecretSigned(obj runtime.Object, numRequeu
 		return true
 	}
 
-	err := v.(OnSignedFn)(secret.Data)
+	certInfo := v.(certInfo)
+
+	// Update certificate expiration time for renewal tracking
+	_ = s.updateCertificateExpiration(secret)
```
tpantelis: We should at least log the returned error here.

pkg/certificate/signing_requestor.go
```diff
@@ -70,8 +90,11 @@ type signingRequestorImpl struct {
 	localClusterID     string
 	localSecretClient  dynamic.ResourceInterface
 	brokerSecretClient dynamic.ResourceInterface
-	onSignedMap        sync.Map
 	keyGenerator       KeyGeneratorFn
+
+	// Certificate management fields
+	renewalTicker *time.Ticker
```
tpantelis: This is unused and not needed.

```suggestion
```

pkg/certificate/signing_requestor.go
```diff
@@ -308,3 +379,72 @@ func deleteIfPresent(ctx context.Context, client dynamic.ResourceInterface, name
...
+
+	logger.Info("Started certificate renewal monitoring", "interval", CertCheckInterval)
+}
+
+// updateCertificateExpiration extracts and stores the certificate expiration time.
+func (s *signingRequestorImpl) updateCertificateExpiration(secret *corev1.Secret) error {
+	certPEM := secret.Data[TLSDataKey]
+	if len(certPEM) == 0 {
+		return errors.Errorf("Failed to find certificate PEM in %q for secret %q", TLSDataKey, secret.Name)
+	}
```
tpantelis: This would be detected by `pem.Decode` below.

```suggestion
```

pkg/certificate/signing_requestor.go
```diff
@@ -308,3 +379,72 @@ func deleteIfPresent(ctx context.Context, client dynamic.ResourceInterface, name
...
+
+	block, _ := pem.Decode(certPEM)
+	if block == nil {
+		return errors.Errorf("Failed to find certificate PEM in %q for secret %q", TLSDataKey, secret.Name)
```
tpantelis: Errors shouldn't be capitalized. Same below.

```suggestion
		return errors.Errorf("failed to find certificate PEM in %q for secret %q", TLSDataKey, secret.Name)
```

tpantelis (changes requested):

pkg/certificate/signing_requestor.go
```diff
@@ -47,15 +55,32 @@ const (
...
 	CADataKey              = "ca.crt"
+
+	// Certificate renewal constants.
+	CertRenewBefore   = 30 * 24 * time.Hour // Renew 30 days before expiration
+	CertCheckInterval = 12 * time.Hour      // Check certificate expiration every 12 hours
+)
+
+var (
+	// CertCheckIntervalForTests can be set to a shorter interval for testing
+	CertCheckIntervalForTests = CertCheckInterval
```
tpantelis: Since the code uses `CertCheckIntervalForTests` you don't really need it, just use `CertCheckInterval` in the tests.

pkg/certificate/signing_requestor.go
```diff
@@ -308,3 +417,125 @@ func deleteIfPresent(ctx context.Context, client dynamic.ResourceInterface, name
...
+	var ips []string
+	for _, ip := range csr.IPAddresses {
+		ips = append(ips, ip.String())
+	}
+
+	return ips, nil
+}
+
+// ipListsEqual compares two IP address lists for equality (order-independent)
+func (s *signingRequestorImpl) ipListsEqual(ips1, ips2 []string) bool {
```
tpantelis: There's a function is admiral that does this: `slices.Equivalent`.

pkg/certificate/signing_requestor.go
```diff
@@ -191,64 +221,106 @@ func (s *signingRequestorImpl) Issue(ctx context.Context, name string, ips []str
...
+
+			keyPEM, csrPEM, err := s.keyGenerator(ips)
+			if err != nil {
+				return nil, errors.Wrapf(err, "error generating key and CSR data for %q and IPs %v", name, ips)
+			}
+
+			secret.Data = map[string][]byte{
+				PrivateKeyDataKey: keyPEM,
+				CSRDataKey:        csrPEM,
+			}
```
tpantelis: We can keep this part above as before and set the fields in `newSecret` as before. Then you don't need `MutateOnCreate`. This is an optimization to avoid generating the key if we don't need to but I think it's negligible in lieu of simplified code.

pkg/certificate/signing_requestor.go
```diff
@@ -191,64 +221,106 @@ func (s *signingRequestorImpl) Issue(ctx context.Context, name string, ips []str
...
+			// We need to extract the IPs from the existing CSR to compare
+			needsNewCSR := false
+
+			if len(existing.Data[CSRDataKey]) == 0 {
+				// No existing CSR data, need to generate
+				needsNewCSR = true
+			} else {
+				// Parse existing CSR to extract IPs and compare
+				existingIPs, err := s.extractIPsFromCSR(existing.Data[CSRDataKey])
+				if err != nil {
```
tpantelis: We shouldn't silence errors, log a warning.

pkg/certificate/signing_requestor.go
```diff
@@ -191,64 +221,106 @@ func (s *signingRequestorImpl) Issue(ctx context.Context, name string, ips []str
...
-				delete(existing.Annotations, RequestSignedLabelKey)
+
+			// Check if we need to regenerate the CSR by comparing the IPs
+			// We need to extract the IPs from the existing CSR to compare
+			needsNewCSR := false
+
+			if len(existing.Data[CSRDataKey]) == 0 {
+				// No existing CSR data, need to generate
+				needsNewCSR = true
+			} else {
```
tpantelis: We don't really need this pre-check as it shouldn't be empty plus parsing the certificate would fail anyway.

pkg/certificate/signing_requestor.go
```diff
@@ -191,64 +221,106 @@ func (s *signingRequestorImpl) Issue(ctx context.Context, name string, ips []str
...
-			existing.Data[PrivateKeyDataKey] = newSecret.Data[PrivateKeyDataKey]
-			existing.Data[CSRDataKey] = newSecret.Data[CSRDataKey]
+			if needsNewCSR {
+				keyPEM, csrPEM, err := s.keyGenerator(ips)
+				if err != nil {
+					return nil, errors.Wrapf(err, "error generating key and CSR data for %q and IPs %v", name, ips)
+				}
+
+				existing.Data[PrivateKeyDataKey] = keyPEM
+				existing.Data[CSRDataKey] = csrPEM
```
tpantelis: As per suggested change above, we can keep this as:

```suggestion
			existing.Data[PrivateKeyDataKey] = newSecret.Data[PrivateKeyDataKey]
			existing.Data[CSRDataKey] = newSecret.Data[CSRDataKey]
```

pkg/certificate/signing_requestor_test.go
```diff
@@ -67,15 +73,12 @@ var _ = Describe("SigningRequestor", func() {
 			Expect(brokerSecret.Data).To(HaveKeyWithValue(certificate.CSRDataKey, localSecret.Data[certificate.CSRDataKey]))
 			Expect(brokerSecret.Data).NotTo(HaveKey(certificate.PrivateKeyDataKey))
 
-			assert.EnsureNoActionsForResource(&t.localDynClient.Fake, "secrets", "update")
```
tpantelis: Why was this removed?

pkg/certificate/signing_requestor_test.go
```diff
@@ -67,15 +73,12 @@ var _ = Describe("SigningRequestor", func() {
...
-
 			Consistently(func() map[string][]byte {
 				return awaitSecret(t.localSecretClient()).Data
 			}).Should(HaveKey(certificate.PrivateKeyDataKey))
 
 			// Re-issuing the same request should be a no-op.
 			Expect(t.signingRequestor.Issue(ctx, secretName, ips, t.onSigned)).To(Succeed())
-			assert.EnsureNoActionsForResource(&t.brokerDynClient.Fake, "secrets", "update")
```
tpantelis: Same here.

pkg/certificate/signing_requestor_test.go
```diff
@@ -373,6 +379,11 @@ func newTestDriver() *testDriver {
 		}
 	})
 
+	AfterEach(func() {
+		// Restore original interval
+		certificate.CertCheckIntervalForTests = certificate.CertCheckInterval
+	})
```
tpantelis: The simpler way to do this is using `DeferCleanup` in `BeforeEach` above on L367:

```
oldCertCheckInterval := certificate.CertCheckInterval
DeferCleanup(func() {
    certificate.CertCheckInterval = oldCertCheckInterval
})

certificate.CertCheckInterval = ...
```

pkg/certificate/signing_requestor_test.go
```diff
@@ -409,6 +420,34 @@ func awaitSecret(client dynamic.ResourceInterface) *corev1.Secret {
 		&corev1.Secret{})
 }
 
+func generateTestCSRForSigningRequestor() []byte {
+	// Generate a test private key
+	privateKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		panic(err)
+	}
```
tpantelis:
```suggestion
	Expect(err).NotTo(HaveOccurred())
```

pkg/certificate/signing_requestor_test.go
```diff
@@ -409,6 +420,34 @@ func awaitSecret(client dynamic.ResourceInterface) *corev1.Secret {
...
+		},
+		SignatureAlgorithm: x509.SHA256WithRSA,
+		IPAddresses:        []net.IP{net.ParseIP("192.168.1.1")},
+	}
+
+	// Create CSR
+	csrDER, err := x509.CreateCertificateRequest(rand.Reader, &template, privateKey)
+	if err != nil {
+		panic(err)
+	}
```
tpantelis:
```suggestion
	Expect(err).NotTo(HaveOccurred())
```

tpantelis (changes requested):

pkg/certificate/signer.go
```diff
@@ -76,6 +105,9 @@ func (s *signerImpl) Start(namespace string) error {
 		return nil
 	}
 
+	// Store the namespace for use in signing operations
+	s.namespace = namespace
```
tpantelis: As mentioned before, the `signerImpl` is intended to handle multiple namespaces so this should not be stored in a field.

pkg/certificate/signer.go
aswinsuryan: changed to in-line

pkg/certificate/signing_requestor_test.go
aswinsuryan: The function seems to be called after the changes, reverted will explore again while fixing UT.

pkg/certificate/signing_requestor.go
```diff
@@ -215,40 +241,64 @@ func (s *signingRequestorImpl) Issue(ctx context.Context, name string, ips []str
...
+			name:      name,
+			ips:       ips,
+			onSigned:  onSigned,
+			expiresAt: time.Time{}, // Will be updated when certificate is signed
+		})
 	}
 
 	if result == util.OperationResultCreated {
-		logger.Infof("Successfully created CSR Secret %q", newSecret.Name)
+		logger.Infof("Successfully created CSR Secret %q", newSecret)
```
tpantelis: This dumps the whole secret which is quite large with the byte data.

```suggestion
		logger.Infof("Successfully created CSR Secret %q", newSecret.Name)
```

pkg/certificate/signing_requestor.go
```diff
@@ -215,40 +241,64 @@ func (s *signingRequestorImpl) Issue(ctx context.Context, name string, ips []str
...
 	} else if result == util.OperationResultUpdated {
-		logger.Infof("Successfully updated CSR Secret %q", newSecret.Name)
+		logger.Infof("Successfully updated CSR Secret %q", newSecret)
 	}
 
-	return errors.Wrapf(err, "error creating or updating CSR Secret %q", newSecret.Name)
+	return errors.Wrapf(err, "error creating or updating CSR Secret %q", newSecret)
```
tpantelis: Same here

pkg/certificate/signing_requestor.go
```diff
@@ -308,3 +399,87 @@ func deleteIfPresent(ctx context.Context, client dynamic.ResourceInterface, name
...
+		}
+
+		// Check if certificate needs renewal
+		timeUntilExpiry := time.Until(info.expiresAt)
+		if timeUntilExpiry <= CertRenewBefore {
+			logger.Infof("Certificate %q expires in %s, renewing", secretName, timeUntilExpiry.String())
+
+			// Renew the certificate by re-issuing it
+			if err := s.Issue(context.TODO(), info.name, info.ips, info.onSigned); err != nil {
+				logger.Errorf(err, "failed to renew certificate %q", secretName)
```
tpantelis:
```suggestion
				logger.Errorf(err, "Failed to renew certificate %q", secretName)
```

pkg/certificate/signing_requestor_test.go
```diff
@@ -409,6 +415,35 @@ func awaitSecret(client dynamic.ResourceInterface) *corev1.Secret {
 		&corev1.Secret{})
 }
 
+func generateTestCSRForSigningRequestor() []byte {
+	// Generate a test private key
+	privateKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		Expect(err).NotTo(HaveOccurred())
+	}
```
tpantelis: No need to check for nil, `Expect` does that.

```suggestion
	Expect(err).NotTo(HaveOccurred())
```

pkg/certificate/signing_requestor_test.go
```diff
@@ -409,6 +415,35 @@ func awaitSecret(client dynamic.ResourceInterface) *corev1.Secret {
...
+		},
+		SignatureAlgorithm: x509.SHA256WithRSA,
+		IPAddresses:        []net.IP{net.ParseIP("192.168.1.1")},
+	}
+
+	// Create CSR
+	csrDER, err := x509.CreateCertificateRequest(rand.Reader, &template, privateKey)
+	if err != nil {
+		Expect(err).NotTo(HaveOccurred())
+	}
```
tpantelis: Same here

pkg/certificate/signer.go
```diff
@@ -132,12 +166,298 @@ func (s *signerImpl) Stop(namespace string) {
...
+			},
+		},
+	}
+
+	result, _, err := util.CreateOrUpdateWithOptions(ctx, util.CreateOrUpdateOptions[*unstructured.Unstructured]{
+		Client: resource.ForDynamic(caSecretClient),
+		Obj:    resource.MustToUnstructured(secret),
+		MutateOnCreate: func(obj *unstructured.Unstructured) (*unstructured.Unstructured, error) {
+			secret := resource.MustFromUnstructured(obj, &corev1.Secret{})
+			err := s.signCASecret(secret, "1")
```
tpantelis:
```suggestion
			err := s.signCASecret(secret, secret.Annotations[CAVersionAnnotation])
```

pkg/certificate/signer.go
```diff
@@ -132,12 +166,298 @@ func (s *signerImpl) Stop(namespace string) {
...
+
+	if result == util.OperationResultCreated {
+		logger.Infof("Successfully created CA Secret %s in namespace %s", secret.Name, namespace)
+	} else if result == util.OperationResultUpdated {
+		logger.Infof("Successfully rotated CA Secret %s in namespace %s", secret.Name, namespace)
+	}
+
+	// Re-sign all existing CSRs with the new CA (handles both creation and rotation cases)
+	if err := s.resignAllCSRs(ctx, namespace); err != nil {
+		logger.Error(err, "failed to re-sign existing CSRs with new CA", "namespace", namespace)
```
tpantelis:
```suggestion
		logger.Error(err, "Failed to re-sign existing CSRs with new CA", "namespace", namespace)
```

pkg/certificate/signer_test.go
```diff
@@ -178,13 +188,45 @@ var _ = Describe("Signer", func() {
...
+		Data: map[string][]byte{certificate.CSRDataKey: csrPEM},
+	}
+}
+
+func generateTestCSR() []byte {
+	// Generate a test private key
+	privateKey, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		panic(err)
+	}
```
tpantelis: Use `Expect`.

pkg/certificate/signer_test.go
```diff
@@ -178,13 +188,45 @@ var _ = Describe("Signer", func() {
...
+			Organization: []string{"submariner.io"},
+		},
+		SignatureAlgorithm: x509.SHA256WithRSA,
+		IPAddresses:        []net.IP{net.ParseIP("192.168.1.1")},
+	}
+
+	// Create CSR
+	csrDER, err := x509.CreateCertificateRequest(rand.Reader, &template, privateKey)
+	if err != nil {
+		panic(err)
```
tpantelis: Use `Expect`.

tpantelis (changes requested):

pkg/certificate/signing_requestor.go
```diff
@@ -215,40 +241,64 @@ func (s *signingRequestorImpl) Issue(ctx context.Context, name string, ips []str
...
+			}
+			// If needsNewCSR is false, we preserve the existing CSR data and signed state
 
 			return resource.MustToUnstructured(existing), nil
 		},
 	})
 	if err == nil {
-		s.onSignedMap.Store(newSecret.Name, onSigned)
+		// Track certificate for renewal (expiration will be set when certificate is signed)
+		s.issuedCerts.Store(newSecret, certInfo{
```
tpantelis:
```suggestion
		s.issuedCerts.Store(newSecret.Name, certInfo{
```

pkg/certificate/signing_requestor.go
```diff
@@ -286,7 +368,16 @@ func (s *signingRequestorImpl) onLocalSecretSigned(obj runtime.Object, numRequeu
 		return true
 	}
 
-	err := v.(OnSignedFn)(secret.Data)
+	certInfo := v.(certInfo)
+
+	// Update certificate expiration time for renewal tracking
+	err := s.updateCertificateExpiration(secret)
+	if err != nil {
+		logger.Warning("Failed to update certificate expiration for secret %q: %v", err, secret.Name)
```
tpantelis:
```suggestion
		logger.Errorf(err, "Failed to update certificate expiration for secret %q", secret.Name)
```

pkg/certificate/signing_requestor.go
```diff
@@ -286,7 +368,16 @@ func (s *signingRequestorImpl) onLocalSecretSigned(obj runtime.Object, numRequeu
...
+		return true
```
tpantelis: I don't think we should return true to retry. An error here means the signed data is invalid which **shouldn't** happen but retrying isn't going to resolve it.

tpantelis (changes requested):

tpantelis (approved):

yboaron (approved):

PR #1196, opened 2025-09-24
Add a package for accessing global configuration

tpantelis:
...initialized from one or more `ConfigMaps`. Consumers call the `Get` function to obtain the configured value for a given key name converted to the desired type or a provided default value if none configured.

The global configuration is utilized by the `zerolog` implementation to determine the `maxVerbosity` for a logger via a name key of the form `log.<prefix>.max-verbosity`.

The global configuration is also utilized by the `workqueue` module for obtaining a `Config`.

### DIFF
--- a/pkg/global/config.go
+++ b/pkg/global/config.go
@@ -1,0 +1,86 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package global
+
+import (
+	"strconv"
+	"sync"
+	"time"
+
+	"github.com/submariner-io/admiral/pkg/log"
+	corev1 "k8s.io/api/core/v1"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+type ValueType interface {
+	string | int | bool | time.Duration
+}
+
+var (
+	configMap sync.Map
+	logger    = log.Logger{Logger: logf.Log.WithName("Global")}
+)
+
+func Init(fromConfigMaps ...*corev1.ConfigMap) {
+	configMap.Clear()
+
+	for _, cm := range fromConfigMaps {
+		if cm == nil {
+			continue
+		}
+
+		for k, v := range cm.Data {
+			configMap.Store(k, v)
+		}
+	}
+}
+
+func Get[T ValueType](name string, defaultValue T) T {
+	v, exists := configMap.Load(name)
+	if !exists {
+		return defaultValue
+	}
+
+	var (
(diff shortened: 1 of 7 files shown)


dfarrell07 (approved):

skitt:
Just a terminology comment: this is a package, not a module :wink:.

tpantelis:
> Just a terminology comment: this is a package, not a module 😉.

yeah I've sometimes used them interchangeably but a module is actually a collection of packages. I stand corrected 😄

pkg/global/config.go
```diff
@@ -0,0 +1,81 @@
...
+	string | int | bool | time.Duration
+}
+
+var configMap sync.Map
+
+func Init(fromConfigMaps ...*corev1.ConfigMap) {
+	configMap.Range(func(key, value interface{}) bool {
+		configMap.Delete(key)
+		return true
+	})
```
skitt:
```suggestion
	configMap.Clear()
```

pkg/global/config.go
```diff
@@ -0,0 +1,81 @@
...
+		returnValue = v
+	case int:
+		returnValue, err = strconv.Atoi(existing)
+	case bool:
+		returnValue, err = strconv.ParseBool(existing)
+	case time.Duration:
+		returnValue, err = time.ParseDuration(existing)
+	}
+
+	utilruntime.Must(errors.Wrapf(err, "Failed to parse config option %q", name))
```
skitt: This creates a denial-of-service risk: since `ConfigMap`s stored in etcd are modifiable by users with sufficient privileges, there is no guarantee that a stored value will have the correct type. Unless I’m missing something, `err` can therefore be non-nil here for reasons other than programmer error.

Instead of panicking, I think the appropriate response here if the value can’t be parsed is to either return an error or log an error and return the default value.

I know the current workqueue config behaviour is to panic on parse errors, but since this is supposed to be more general-purpose I don’t think that’s appropriate.

skitt (changes requested):

pkg/global/config.go
tpantelis: Good point. I'll log an error and return the default.

pkg/global/config.go
```diff
@@ -0,0 +1,88 @@
...
+	case int:
+		returnValue, err = strconv.Atoi(existing)
+	case bool:
+		returnValue, err = strconv.ParseBool(existing)
+	case time.Duration:
+		returnValue, err = time.ParseDuration(existing)
+	}
+
+	if err != nil {
+		logger.Errorf(err, "Failed to parse config option %q - using default value %v", name, defaultValue)
```
skitt:
```suggestion
		logger.Errorf(err, "Failed to parse config option %q value %q - using default value %v", name, existing, defaultValue)
```

pkg/global/config.go
skitt: This one is still left over from previous iterations (use `configMap.Clear()` instead).

skitt (changes requested):

pkg/global/config.go
tpantelis: There is not `Clear` method on `sync.Map`.

skitt: https://pkg.go.dev/sync#Map.Clear

tpantelis: I didn't see it b/c my GoLand was using Go 1.22

tpantelis:
The unit test failure isn't related and is addressed by https://github.com/submariner-io/admiral/pull/1198.

aswinsuryan (approved):

skitt (approved):

PR #1200, opened 2025-09-27
Add configmap.WatchAndSignalOnChange function

tpantelis:
...that watches specified `ConfigMaps` and sends the specified signal when a `ConfigMap` is created/updated/deleted. The intended use is to send a SIGINT to restart the process for the `ConfigMap` change to take effect.

### DIFF
--- a/pkg/configmap/config_map_test.go
+++ b/pkg/configmap/config_map_test.go
@@ -1,0 +1,178 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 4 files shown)


pkg/configmap/config_map_test.go
```diff
@@ -0,0 +1,180 @@
...
+	BeforeEach(func() {
+		k8sClient = k8sfake.NewClientset()
+
+		var cancel context.CancelFunc
+
+		ctx, cancel = context.WithCancel(context.Background()) //nolint:fatcontext // Ignore
+
+		DeferCleanup(func() {
+			cancel()
+		})
```
skitt:
```suggestion
		DeferCleanup(cancel)
```

:wink:

skitt (approved):
See also https://github.com/kubernetes/kubernetes/issues/22368 (no solution there, just as an indicator that this is a shared use case).

aswinsuryan (approved):

PR #1229, opened 2025-11-13
Add additional fake K8s client support

tpantelis:
See commits for details.

### DIFF
--- a/pkg/fake/interceptors.go
+++ b/pkg/fake/interceptors.go
@@ -1,0 +1,90 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package fake
+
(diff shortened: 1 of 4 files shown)


pkg/fake/interceptors.go
```diff
@@ -0,0 +1,90 @@
...
+			opts ...ctrlClient.GetOption,
+		) error {
+			if reflect.TypeOf(obj) == reflect.TypeOf(ofType) {
+				return errors.New("mock Get error")
+			}
+
+			return client.Get(ctx, key, obj, opts...)
+		},
+	}
+}
```
skitt: Whether this is useful or not depends on what the interceptor declarations end up like, but you could do

```suggestion
func FailingGetInterceptor[T ctrlClient.Object](_ T) interceptor.Funcs {
	return interceptor.Funcs{
		Get: func(ctx context.Context, client ctrlClient.WithWatch, key ctrlClient.ObjectKey, obj ctrlClient.Object,
			opts ...ctrlClient.GetOption,
		) error {
			if reflect.TypeOf(obj) == reflect.TypeFor[T]() {
				return errors.New("mock Get error")
			}

			return client.Get(ctx, key, obj, opts...)
		},
	}
}
```

and even drop the argument entirely if it only ends up being a type-carrying instance. That is to say, if this ends up being called as `FailingGetInterceptor(package.Foo{})`, then instantiation can be avoided with `FailingGetInterceptor[package.Foo]()`.

Similar transforms can be applied to the other interceptor functions.

skitt (approved):

pkg/fake/interceptors.go
tpantelis: yeah - I think that might work - I'll have to go look at the subctl code I have local that uses it (it's been a couple weeks since I did this).

tpantelis: That worked nicely! 🥂

PR #1231, opened 2025-11-19
Enhance test reporter assertions

tpantelis:
Added a method to assert a failure message exists containing an arbitrary number of strings. Added similar method for a warning message.

### DIFF
--- a/pkg/reporter/test/tracker.go
+++ b/pkg/reporter/test/tracker.go
@@ -58,6 +58,25 @@ func (t *Tracker) AssertContainsFailure(matcher types.GomegaMatcher) {
 	Expect(t.failures).To(ContainElement(matcher))
 }
 
+// AssertFailureContainsStrings checks if any failure message contains all the given strings. This a
+// convenience function equivalent to
+//
+//	AssertContainsFailure(And(ContainSubstring("1"), ContainSubstring("2"), ...))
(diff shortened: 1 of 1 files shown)


pkg/reporter/test/tracker.go
```diff
@@ -58,6 +58,19 @@ func (t *Tracker) AssertContainsFailure(matcher types.GomegaMatcher) {
 	Expect(t.failures).To(ContainElement(matcher))
 }
 
+func (t *Tracker) AssertFailureContainsStrings(s ...string) {
+	if len(s) == 0 {
+		t.AssertHasFailure()
```
skitt, line 69: The semantics aren’t obvious to me — when would `AssertFailureContainsStrings()` be called with no arguments, and why should it fail?

tpantelis: I'm using this generically where the caller of this function is passed the variadic strings. The semantics in my usage is a failure is expected but, in some cases, the caller may not care about a specific message. I thought these semantics might be generally useful but perhaps not.

skitt: That’s fine, it just needs a doc comment to make it explicit :wink: (same for the other function).

skitt (approved):

Issue #1271, opened 2026-02-12
Support for controlling Local/Remote syncer QPS and Burst


msft-paddy14:
When syncing resources between the local cluster and broker, the broker syncer component can experience client-side throttling with warnings.

This happens because the default client-go rate limits (QPS=5, Burst=10) are too low for high-throughput scenarios with many ServiceImports or EndpointSlices from lighthouse agent perspecitve.

Currently, the only way to configure these values is to modify the code (Lighthouse separately also creates a  client dependency during agent setup which is not in scope for this issue)

Proposed solution:
Add support for configuring client-go QPS and Burst settings via environment variables.

SUBMARINER_CLIENT_K8S_QPS - Maximum queries per second to the API server (default: 5)
SUBMARINER_CLIENT_K8S_BURST - Maximum burst for throttle to the API server (default: 10)

These environment variables apply to both the local cluster client and the broker cluster client, providing a single configuration point for all API server communication from broker.

tpantelis:
We could use the `submariner-global` `ConfigMap` to store these settings and have the operator read the `ConfigMap` and propagate the settings set via environment variables to the deployments and daemonsets.

msft-paddy14:
Makes sense. So these changes would be done in operator repo after we expose the functionality here in admiral, correct?

Also this can only be done during bootstrap I guess, updating environment variables by patching configmap would still need a restart of pods as we only use these during initial broker client setup

tpantelis:
> Also this can only be done during bootstrap I guess, updating environment variables by patching configmap would still need a restart of pods as we only use these during initial broker client setup

The operator would update the env vars in the deployments which would restart the pods.

tpantelis:
There's a couple approaches to this. As mentioned above, have the operator read the settings from a `ConfigMap` and propagate to deployments/daemonsets via env vars. However, we've been moving away from  passing args vis env vars.

I think we should have the component pods read the settings from a `ConfigMap`. Most already do this for other settings, typically read from a component-specific `ConfigMap`.  For this we could use a shared `ConfigMap`. We already have one defined named `submariner-global` that the [route-agent](https://github.com/submariner-io/submariner/blob/6c12b78a91f1a29b88e7bcc3838ba2cbc7d73eed/pkg/routeagent_driver/main.go#L123) uses. Usage is:

```
import "github.com/submariner-io/admiral/pkg/configmap"
import "github.com/submariner-io/admiral/pkg/global"
...

// Read the ConfigMap
[...]

PR #1272, opened 2026-02-12
Add support for broker client-go qps and burst config

msft-paddy14:
Adds support for configuring client-go QPS and Burst settings via environment variables for the broker syncer component.
Addresses: #1271 
When syncing resources between the local cluster and broker, the broker syncer can experience client-side throttling.

The environment variables apply to both the local cluster client and the broker cluster client created by the broker syncer in ensureClients().

Add support for configuring client-go QPS and Burst via environment variables:
- BROKER_K8S_QPS: Maximum queries per second to the API server (default: 5)
- BROKER_K8S_BURST: Maximum burst for throttle to the API server (default: 10)

## Summary by CodeRabbit

* **New Features**
  * Broker client request pacing via QPS/Burst configurable (explicit broker settings take precedence).
  * New configuration keys added for Kubernetes client QPS/Burst (including broker-specific keys).

* **Tests**
  * Added tests validating QPS/Burst propagation, precedence between broker and global config, and defaults.

* **Bug Fixes**
  * Clarified error message to reference broker environment configuration.


pkg/syncer/broker/syncer.go
```diff
@@ -361,6 +361,15 @@ func (c *SyncerConfig) createBrokerClient() error {
...
 	var err error
 
+	// Get client QPS/Burst settings from environment variables
+	brokerSpec, err := getBrokerSpecification()
+	if err != nil {
+		return err
+	}
+
+	// Apply QPS and Burst settings to local REST config
+	applyQPSBurst(c.LocalRestConfig, brokerSpec)
```
tpantelis: The caller creates and provides the `LocalRestConfig` so the caller should configure the rate limiter settings and this component shouldn't set them as a side effect. Same with the `BrokerRestConfig`, if provided by the caller.

This component should only configure  the rate limiter settings for the `BrokerRestConfig` if it creates it via the env variables (ie in `createBrokerClient` after line 345).

tpantelis (changes requested):

pkg/syncer/broker/syncer.go
msft-paddy14: Makes sense. Addressed the changes.

pkg/syncer/broker/config.go
```diff
@@ -31,9 +31,11 @@ type brokerSpecification struct {
 	APIServer       string
 	APIServerToken  string
 	RemoteNamespace string
-	Insecure        bool `default:"false"`
 	Ca              string
 	Secret          string
+	Burst           int     `default:"10"`
+	QPS             float32 `default:"5"`
```
tpantelis: I don't think we should default these, ie let K8s default them.

pkg/syncer/broker/syncer_test.go
```diff
@@ -688,4 +698,49 @@ var _ = Describe("Broker Syncer", func() {
...
+			os.Setenv("BROKER_K8S_REMOTENAMESPACE", test.RemoteNamespace)
+		})
+
+		It("should apply default QPS (5) and Burst (10) to broker rest config before client creation", func() {
+			Expect(brokerQPSAtCreation).To(Equal(float32(5)))
+			Expect(brokerBurstAtCreation).To(Equal(10))
+		})
+	})
+
+	When("BrokerRestConfig is explicitly provided with custom QPS/Burst", func() {
```
tpantelis: There's already a test spec that configures the `BrokerRestConfig` on line 638 so we can incorporate this into that one (ie do the verification in the It on line 649).

pkg/syncer/broker/syncer_test.go
```diff
@@ -688,4 +698,49 @@ var _ = Describe("Broker Syncer", func() {
 			})
 		})
 	})
+
+	When("client QPS/Burst environment vars are specified", func() {
```
tpantelis: This can be added as a spec under line 665 above:

```
When("broker config environment vars are specified", func() {
    ...
    Context("including QPS and Burst", func() {
		BeforeEach(func() {
			os.Setenv("BROKER_K8S_QPS", "100")
			os.Setenv("BROKER_K8S_BURST", "500")
		})

		It("should apply the QPS and Burst settings to the broker rest config", func() {
			...
		})
	})
})
```

pkg/syncer/broker/syncer_test.go
```diff
@@ -688,4 +698,49 @@ var _ = Describe("Broker Syncer", func() {
...
+		BeforeEach(func() {
+			os.Setenv("BROKER_K8S_APISERVER", "broker-host")
+			os.Setenv("BROKER_K8S_REMOTENAMESPACE", test.RemoteNamespace)
+			os.Setenv("BROKER_K8S_QPS", "100")
+			os.Setenv("BROKER_K8S_BURST", "500")
+		})
+
+		It("should apply QPS and Burst to broker rest config before client creation", func() {
+			Expect(brokerQPSAtCreation).To(Equal(float32(100)))
+			Expect(brokerBurstAtCreation).To(Equal(500))
```
tpantelis: The broker `rest.Config` is already captured in `actualBrokerRestConfig` so you can use that to verify as is done on line 676 (ie `brokerQPSAtCreation` and `brokerBurstAtCreation` aren't needed).

pkg/syncer/broker/syncer_test.go
```diff
@@ -688,4 +698,49 @@ var _ = Describe("Broker Syncer", func() {
...
+		})
+	})
+
+	When("client QPS/Burst environment vars are not specified", func() {
```
tpantelis: This can be incorporated into the test spec on line 675 above that already tests broker en vars, ie verify QPS and Burst aren't set.

tpantelis (changes requested):

pkg/syncer/broker/syncer.go
```diff
@@ -343,6 +342,7 @@ func (c *SyncerConfig) createBrokerClient() error {
 			c.BrokerRestConfig, authorized, err = resource.GetAuthorizedRestConfigFromData(spec.APIServer, spec.APIServerToken, spec.Ca,
 				&rest.TLSClientConfig{Insecure: spec.Insecure}, *gvr, spec.RemoteNamespace)
 		}
+		applyQPSBurst(c.BrokerRestConfig, spec)
```
tpantelis: Please see my comment https://github.com/submariner-io/admiral/issues/1271#issuecomment-3893451988 in the related issue. The settings would be obtained from `global.Get` instead of the `brokerSpecification` via env vars.

pkg/syncer/broker/config.go
msft-paddy14: Removed default values. Also made the check to !=0 for when settings these, since it can be -1 to remove limits

pkg/syncer/broker/syncer_test.go
msft-paddy14: Yeah. removed. Had misinterpreted the coderrabbit feedback

pkg/syncer/broker/syncer_test.go
msft-paddy14: moved

pkg/syncer/broker/syncer_test.go
msft-paddy14: moved

pkg/syncer/broker/syncer.go
```diff
@@ -384,6 +385,20 @@ func (c *SyncerConfig) ensureClients() error {
 	return nil
 }
 
+func applyQPSBurst(restConfig *rest.Config) {
+	if restConfig == nil {
+		return
+	}
+
+	if qps := global.Get(global.K8sBrokerClientQPS, 0); qps != 0 {
```
tpantelis: Minor thing but if the value was negative, we would silently ignore it here. We could just let K8s validate it and return an error - the pod would terminate but at least it would be reported to the user so they're aware of it.

pkg/syncer/broker/syncer_test.go
```diff
@@ -74,8 +76,12 @@ var _ = Describe("Broker Syncer", func() {
 		os.Unsetenv("BROKER_K8S_REMOTENAMESPACE")
 		os.Unsetenv("BROKER_K8S_INSECURE")
 		os.Unsetenv("BROKER_K8S_SECRET")
+		os.Unsetenv("BROKER_K8S_QPS")
+		os.Unsetenv("BROKER_K8S_BURST")
```
tpantelis: Unused now.


```suggestion
```

tpantelis (changes requested):

pkg/syncer/broker/syncer.go
msft-paddy14: qps != 0 should still set negative values. If the check was > 0, then it was a problem, right?

tpantelis: yeah it's fine. According to the K8s doc, a negative value is actually valid. But we really don't need any check here, ie just do

`restConfig.QPS = float32(global.Get(global.K8sBrokerClientQPS, 0))`

There's actually no code path where line 394 below can be reached so let's remove it.

Also, this is only called on line 345 so we might as well inline it:

```
if c.BrokerRestConfig != nil {
    c.BrokerRestConfig.QPS = float32(global.Get(global.K8sBrokerClientQPS, 0))
    c.BrokerRestConfig.Burst = global.K8sBrokerClientBurst, 0)
}
```

msft-paddy14: inlined as per suggestions

tpantelis (approved):

PR #1293, opened 2026-02-19
Add configurable number of workers for workqueues

### LINKED ISSUE #1292: Support configurable worker count for workqueue  to improve scalability + queue metrics to measure this
The current workqueue implementation uses a fixed number of workers limiting scalability for high number of EndpointSlice events. Based on performance testing, we've identified that increasing the worker count significantly improves sync latency.

### Current issue

Today, a broker syncer takes a config with ResourceConfig creating a remote and local syncer for each individual resource. However, the queue for each of these runs a **single worker** today.

https://github.com/submariner-io/admiral/blob/4ba4e26833545ab75765c35731d8ac18acc5e559/pkg/workqueue/queue.go#L134

Ideally, using ResourceConfig label selectors, consumers of admiral can still use different resources and have different queues but not handle high volume traffic for a given resource.

### Test Setup

Used shipyard setup to create 3 clusters - 1 broker and 2 participating clusters (ClusterB and ClusterC).

Simplified test workload:
1. Create N endpointslices in ClusterB, also each endpointslice here only contained single entry for endpoint
2. Create ServiceExport in ClusterB and kick off latency measurement
3. Await all endpointslices in ClusterC and record latency on completion

### Results

| EPS Count | 1 Worker | 5 Workers | 50 Workers | 250 Workers | Speedup (1→250) |
|-----------|----------|-----------|------------|-------------|-----------------|
| 1         | 521ms    | 185ms     | 13-15ms    | 14ms        | **~37x**        |
| 100       | 2.0s     | 4.9s      | 22-29ms    | 23ms        | **~87x**        |
| 500       | 12.5s    | 18.8s     | 1.6-2.1s   | 2.1s        | **~6x**         |
| 1000      | 32.6s    | 22.6s     | 3.7-7.7s   | 3.7s        | **~9x**         |
| 2500      | 1m7s     | 37.7s     | 17.8-25.3s | 19.3s       | **~3.5x**       |
| 5000      | **2m1s** | 1m27s     | 36-56s     | **30.9s**   | **~4x**         |
| 10000     | N/A      | N/A       | 1m50-2m    | **1m24s**   | -               |

### Additional changes for testing

Additional changes for testing which might impact perf for others:
1. Increased throttling limits for broker/local/workqueue
2. Added debug logs for QueueLength to measure lag
3. Minor federation logic change to prevent list calls for existence check during Create event

With the above data, we do see performance scale with workers and giving the clients of Admiral an option would be great to have. Note the gains would be both for ClusterB→Broker sync and Broker←ClusterC sync.

Additionally, we can consider adding logging for QueueLength to measure lag so clients can decide on the value based on it. Based on logging, I saw QueueLength decrease rapidly from **~10k → ~4k → ~15-20** as I increased worker count.

### Proposed solution

Add support for configurable worker count in the workqueue configuration:
https://github.com/submariner-io/admiral/blob/4ba4e26833545ab75765c35731d8ac18acc5e559/pkg/workqueue/config.go#L37

### Few tricky cases to handle

1. Based on above param, the queue Run method can create corresponding workers. Consumer code in lighthouse will probably need to expose more verbose settings as it uses multiple syncers for service import controller but only single for endpointslice controller. Creating same workers for both wouldn't make sense.

2. Race with multiple goroutines: For queue Get/Enqueue, k8s already handles concurrent access. However, admiral code is using a wrapper priority queue for priority setting for each entry called. This might need additional protection (but I assume it's not that critical since it's only used to mark low priority items during resync as per my understanding).

tpantelis:
Introduce `NumWorkers` configuration to allow multiple concurrent workers processing items from the same workqueue, improving throughput for parallel processing scenarios.

Fixes https://github.com/submariner-io/admiral/issues/1292

## Release Notes

* **New Features**
  * Work queue configuration now supports configurable concurrent workers, enabling parallel processing of tasks instead of sequential execution.

* **Tests**
  * Added comprehensive tests validating concurrent worker behavior, including verification that all items are processed correctly and concurrent execution operates within configured worker limits.

### DIFF
--- a/pkg/workqueue/config.go
+++ b/pkg/workqueue/config.go
@@ -32,6 +32,7 @@ const (
 	BucketRateLimiterItemsPerSecKey = "bucket-rate-limiter-items-per-sec"
 	BucketRateLimiterMaxBurstKey    = "bucket-rate-limiter-max-burst"
 	MaxVerbosityKey                 = "max-verbosity"
+	NumWorkersKey                   = "num-workers"
 )
 
 type Config struct {
@@ -41,6 +42,7 @@ type Config struct {
 	BucketRateLimiterItemsPerSec int
 	BucketRateLimiterMaxBurst    int
 	MaxVerbosity                 int
+	NumWorkers                   int
 }
 
 func DefaultConfig() Config {
@@ -50,6 +52,7 @@ func DefaultConfig() Config {
 		OverallRateLimiterMaxDelay:   5 * time.Minute,
 		BucketRateLimiterItemsPerSec: 10,
 		BucketRateLimiterMaxBurst:    500,
+		NumWorkers:                   1,
 	}
 }
 
@@ -74,6 +77,7 @@ func ConfigFromGlobal(keyPrefix string, defaultConfig *Config) *Config {
 	config.BucketRateLimiterMaxBurst = global.Get(ToConfigMapDataKey(keyPrefix, BucketRateLimiterMaxBurstKey),
 		config.BucketRateLimiterMaxBurst)
 	config.MaxVerbosity = global.Get(ToConfigMapDataKey(keyPrefix, MaxVerbosityKey), config.MaxVerbosity)
+	config.NumWorkers = global.Get(ToConfigMapDataKey(keyPrefix, NumWorkersKey), config.NumWorkers)
 
 	return &config
 }
--- a/pkg/workqueue/queue_test.go
+++ b/pkg/workqueue/queue_test.go
@@ -23,6 +23,7 @@ import (
 	"errors"
 	"strconv"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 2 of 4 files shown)


pkg/workqueue/queue_test.go
```diff
@@ -289,6 +289,87 @@ var _ = Describe("Work Queue", func() {
...
+				})
+
+				g.Expect(count).To(Equal(itemCount))
+			}).Within(5 * time.Second).Should(Succeed())
+
+			// Verify that multiple workers were active concurrently
+			Expect(maxConcurrent).To(BeNumerically(">", 1), "Expected multiple workers to be active concurrently")
+			Expect(maxConcurrent).To(BeNumerically("<=", 5), "Expected at most 5 workers to be active concurrently")
+		})
+	})
```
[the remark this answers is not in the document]
tpantelis: This was addressed using atomics.  Please re-review and resolve.

tpantelis:
@coderabbitai approve

pkg/workqueue/config.go
```diff
@@ -50,6 +52,7 @@ func DefaultConfig() Config {
 		OverallRateLimiterMaxDelay:   5 * time.Minute,
 		BucketRateLimiterItemsPerSec: 10,
 		BucketRateLimiterMaxBurst:    500,
+		NumWorkers:                   1,
```
skitt, line 55: Would it make sense to have this default to the number of available logical CPUS?


```suggestion
		NumWorkers:                   runtime.NumCPU(),
```

This would give more headroom by default.

tpantelis: I think we should leave it up to the client whether more than one should be used since there may be concurrency concerns in the client's process function (see doc comment on the `Run` methoc).

skitt (approved):

tpantelis:
@coderabbitai approve

PR #1295, opened 2026-02-20
Add syncer metrics for federation, transform and queue wait time

### LINKED ISSUE #1294: Metrics support for federation, transform, and queue operations in Resource Syncer
Admiral Resource Syncer currently only exposes a SyncCounter gauge that tracks the number of create/update/delete operations. During benchmarking of Lighthouse with high throughput workloads, this single counter is insufficient to diagnose performance bottlenecks in the sync pipeline. 
There are different phases which needs breakdown to pin point latency issues better.
1. QueueLatencyMs/MaxQueueLength - time spent in queue by an event before being processed. This can help identify delay in processing and need to increase workers
2. FederationLatencyMs- Calls to API server can be slow (e.g. due to API server throttling or network latency), there's no metric to surface this.
3. TransformationLatencyMs - though generally negligible, it's good to surface this for future changes

### Sample Metrics based on testing:

Image

Image

### Proposed solution:

Matching existing SyncCounter/SyncCounterOpts pattern: Callers can either pass *Counter/*Opts for similar pattern for these metrics.

**Queue length as Histogram (not Gauge)**:  instantaneous value at scrape time which might be 0 since the queue drains between scrapes. A Histogram records the distribution of queue depths at every enqueue so we can get Max

**Queue wait tracking:** Enqueue timestamps are stored in a sync.Map (keyed by workqueue key) before Enqueue() is called, and consumed on dequeue via LoadAndDelete

Instrument transformation and federation phases similarly.

msft-paddy14:
Adds optional Prometheus histogram metrics to ResourceSyncer for observing federation latency, transform latency, queue wait time, and queue depth distribution.
Fixes #1294 

## Summary by CodeRabbit

* **New Features**
  * Centralized, configurable metrics subsystem for syncers with optional custom collectors and default buckets.
  * Instrumentation added for enqueue/dequeue, transform, federation durations, queue wait time, queue length, and per-operation counters.
  * Work‑queue length query exposed for observability.
  * Configurable max log verbosity added for syncer components.

* **Tests**
  * Tests updated to initialize and validate the new metrics configuration and collectors.

### DIFF
--- a/pkg/syncer/broker/syncer.go
+++ b/pkg/syncer/broker/syncer.go
@@ -25,7 +25,6 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-	"github.com/prometheus/client_golang/prometheus"
 	"github.com/submariner-io/admiral/pkg/federate"
 	"github.com/submariner-io/admiral/pkg/global"
 	"github.com/submariner-io/admiral/pkg/log"
@@ -111,8 +110,8 @@ type ResourceConfig struct {
 	// BrokerWorkQueueConfig if specified, configures the underlying work queue for processing broker resources.
 	BrokerWorkQueueConfig *workqueue.Config
 
-	// SyncCounterOpts used to pass name and help text to resource syncer Gauge
-	SyncCounterOpts *prometheus.GaugeOpts
+	// Metrics if specified, configures optional Prometheus metrics for federation, transform, and queue operations.
+	Metrics syncer.MetricsConfig
 }
 
 type SyncerConfig struct {
@@ -214,18 +213,7 @@ func NewSyncer(config SyncerConfig) (*Syncer, error) { //nolint:gocritic // Mini
 	for i := range config.ResourceConfigs {
 		rc := &config.ResourceConfigs[i]
 
-		var syncCounter *prometheus.GaugeVec
-		if rc.SyncCounterOpts != nil {
-			syncCounter = prometheus.NewGaugeVec(
-				*rc.SyncCounterOpts,
-				[]string{
-					syncer.DirectionLabel,
-					syncer.OperationLabel,
-					syncer.SyncerNameLabel,
-				},
-			)
-			prometheus.MustRegister(syncCounter)
-		}
+		metrics := syncer.ResolveMetricsConfig(rc.Metrics)
 
 		federator := rc.BrokerFederator
 		if federator == nil {
@@ -255,7 +243,7 @@ func NewSyncer(config SyncerConfig) (*Syncer, error) { //nolint:gocritic // Mini
 			WorkQueueConfig:     rc.LocalWorkQueueConfig,
 			Scheme:              config.Scheme,
 			ResyncPeriod:        rc.LocalResyncPeriod,
-			SyncCounter:         syncCounter,
+			Metrics:             metrics,
 			MaxLogVerbosity:     config.MaxLogVerbosity,
 		})
 		if err != nil {
@@ -295,7 +283,7 @@ func NewSyncer(config SyncerConfig) (*Syncer, error) { //nolint:gocritic // Mini
 			WorkQueueConfig:     rc.BrokerWorkQueueConfig,
 			Scheme:              config.Scheme,
 			ResyncPeriod:        rc.BrokerResyncPeriod,
-			SyncCounter:         syncCounter,
+			Metrics:             metrics,
 			NamespaceInformer:   config.NamespaceInformer,
 			MaxLogVerbosity:     config.MaxLogVerbosity,
 		})
--- a/pkg/syncer/metrics.go
+++ b/pkg/syncer/metrics.go
@@ -1,0 +1,250 @@
+/*
+SPDX-License-Identifier: Apache-2.0
+
+Copyright Contributors to the Submariner project.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package syncer
+
+import (
+	"sync"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// MetricsConfig configures optional Prometheus metrics for syncer operations.
+// For each metric, either pass Opts to let the syncer create and register it,
+// or pass a pre-created collector to share it across syncers.
+//
+// When a MetricsConfig is shared between multiple syncers (e.g. the local and
+// remote syncers created by broker/syncer.go), callers should pre-resolve
+// once and pass the live collectors.
+type MetricsConfig struct {
+	// SyncCounterOpts if specified, used to create a gauge to record sync counter metrics.
+	// Alternatively the gauge can be created directly and passed via SyncCounter,
+	// in which case SyncCounterOpts is ignored.
+	SyncCounterOpts *prometheus.GaugeOpts
+
+	// SyncCounter if specified, used to record counter metrics.
+	SyncCounter *prometheus.GaugeVec
+
+	// Federation duration in ms (Distribute/Delete).
+	FederationDurationMsOpts *prometheus.HistogramOpts
+	FederationDurationMs     *prometheus.HistogramVec
+
+	// Transform function duration in ms.
+	TransformDurationMsOpts *prometheus.HistogramOpts
+	TransformDurationMs     *prometheus.HistogramVec
+
+	// Time items spend waiting in the work queue, in ms.
+	QueueWaitDurationMsOpts *prometheus.HistogramOpts
+	QueueWaitDurationMs     *prometheus.HistogramVec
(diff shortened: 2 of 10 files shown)


pkg/syncer/metrics.go
```diff
@@ -0,0 +1,220 @@
...
+	transformDurationMs  *prometheus.HistogramVec
+	queueWaitDurationMs  *prometheus.HistogramVec
+	queueLength          *prometheus.HistogramVec
+	enqueueTimes         sync.Map
+
+	// Non-nil when we created the collector ourselves (from Opts) and own its lifecycle.
+	federationDurationMsOpts *prometheus.HistogramOpts
+	transformDurationMsOpts  *prometheus.HistogramOpts
+	queueWaitDurationMsOpts  *prometheus.HistogramOpts
+	queueLengthOpts          *prometheus.HistogramOpts
```
tpantelis: To simplify, we can define a `MetricsConfig` field rather than duplicating fields.


```suggestion
	MetricsConfig
	enqueueTimes sync.Map
```

pkg/syncer/broker/syncer.go
```diff
@@ -309,6 +316,68 @@ func NewSyncer(config SyncerConfig) (*Syncer, error) { //nolint:gocritic // Mini
 	return brokerSyncer, nil
 }
 
+// resolveMetricsConfig pre-resolves any Opts-based fields in the MetricsConfig
+// into live HistogramVec collectors. This mirrors the SyncCounterOpts pattern:
+// the collector is created and registered once here, and the live collector is
+// passed to both the local and remote syncers so they never attempt duplicate
+// registration.
+func resolveMetricsConfig(cfg *syncer.MetricsConfig) *syncer.MetricsConfig {
```
tpantelis: The code in this function is pretty much duplicated with `syncer.newSyncerMetrics`.  It seems like we should be able to move this function to _metrics.go_  (and make it public) and use both here and in `syncer.newSyncerMetrics`.

pkg/syncer/types.go
```diff
@@ -177,6 +177,9 @@ type ResourceSyncerConfig struct {
 	// WorkQueueConfig if specified, configures the underlying work queue
 	WorkQueueConfig *workqueue.Config
 
+	// Metrics if specified, configures optional Prometheus metrics for federation, transform, and queue operations.
+	Metrics *MetricsConfig
```
tpantelis: Since `MetricsConfig` contains all pointer fields, this really doesn't need to be a pointer field. Not a big deal but it would simplify code a bit not having to deal with another possibly nil pointer.


```suggestion
	Metrics MetricsConfig
```

tpantelis (changes requested):
It would nice to refactor `SyncCounter`/`SyncCounterOpts` into `MetricsConfig` but that can be done in a separate PR if you prefer..

msft-paddy14:
> It would nice to refactor `SyncCounter`/`SyncCounterOpts` into `MetricsConfig` but that can be done in a separate PR if you prefer..

I thought about doing this but it's exposed as a field in ResourceConfig and lighthouse agent is already consuming it. Would be a breaking change for the upgrade to new version then, right? I do agree it will be cleaner to have everything together though. Can do it in a separate PR to avoid issues as you suggested

tpantelis:
> > It would nice to refactor `SyncCounter`/`SyncCounterOpts` into `MetricsConfig` but that can be done in a separate PR if you prefer..
> 
> I thought about doing this but it's exposed as a field in ResourceConfig and lighthouse agent is already consuming it. Would be a breaking change for the upgrade to new version then, right? 

yeah that's fine - we can adjust lighthouse code when we bump the Go dependency.

pkg/syncer/types.go
msft-paddy14: Have converted to value instead of pointer. Still kept pointer in functions as there were golanggci-lint errors for too large objects being coped.

pkg/syncer/metrics.go
msft-paddy14: I initially thought it might create problems if we have internal metrics since MetricsConfig is public but you're right we can always use another field in syncerMetrics for that, so reusing MetricsConfig type here makes sense

pkg/syncer/metrics.go
```diff
@@ -0,0 +1,259 @@
...
+
+	// Queue depth sampled on each enqueue/dequeue. Histogram captures the
+	// distribution so you can query max via histogram_quantile(1.0, ...).
+	QueueLengthOpts *prometheus.HistogramOpts
+	QueueLength     *prometheus.HistogramVec
+}
+
+var (
+	DefaultDurationMsBuckets  = []float64{1, 5, 10, 25, 50, 100, 250, 500, 1000, 2500, 5000}
+	DefaultQueueLengthBuckets = []float64{0, 1, 2, 3, 5, 8, 10, 15, 20, 50, 100, 500}
```
tpantelis: These no longer need to be public.

pkg/syncer/resource_syncer_test.go
```diff
@@ -72,7 +72,7 @@ func testLocalToRemote() {
 	d := newTestDriver(test.LocalNamespace, "", syncer.LocalToRemote)
 
 	BeforeEach(func() {
-		d.config.SyncCounterOpts = &prometheus.GaugeOpts{
+		d.config.Metrics.SyncCounterOpts = &prometheus.GaugeOpts{
```
tpantelis: To cover the new metrics:


```suggestion
		d.config.Metrics = syncer.MetricsConfig{
			SyncCounterOpts: &prometheus.GaugeOpts{
				Name: "sync-counter",
			},
			FederationDurationMsOpts: &prometheus.HistogramOpts{
				Name: "federation",
			},
			QueueWaitDurationMsOpts: &prometheus.HistogramOpts{
				Name: "queue-wait",
			},
			QueueLengthOpts: &prometheus.HistogramOpts{
				Name: "queue-length",
			},
		}
```

To cover `TransformDurationMs`, add

```
t.config.Metrics = syncer.MetricsConfig{
    TransformDurationMsOpts: &prometheus.HistogramOpts{
	    Name: "transform",
	},
}
```

to `BeforeEach` in `newTransformFunctionTestDriver`.

pkg/syncer/metrics.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	if original.QueueLengthOpts != nil && original.QueueLength == nil {
+		resolved.QueueLengthOpts = original.QueueLengthOpts
+	}
+}
+
+func (m *syncerMetrics) unregister() {
+	if m == nil {
+		return
```
tpantelis: This line is never reached in unit tests because `syncerMetrics` is never nil. As such, I think we should remove the `m == nil` checks here and in other methods. My philosophy generally is that if it's impossible to devise a test to cover a line then the line probably shouldn't be there.

pkg/syncer/metrics.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+// ResolveMetricsConfig pre-resolves any Opts-based fields in the MetricsConfig
+// into live HistogramVec collectors. The returned config has live collectors set
+// and Opts fields niled out. This is useful when sharing a MetricsConfig between
+// multiple syncers (e.g. broker local and remote syncers) to avoid duplicate
+// registration.
+func ResolveMetricsConfig(cfg *MetricsConfig) MetricsConfig {
+	if cfg == nil {
+		return MetricsConfig{}
```
tpantelis: All call sites pass in a valid pointer so this line is not covered by unit tests. Options:

1) Add a unit test that specifically calls this with nil
2) Don't check the pointer and assume the caller passes non-nil
3) Change `cfg` param to a non-pointer value


I think 1 is overkill. 2 is reasonable but I prefer 3. The `hugeParam` linter will complain but this function is used in setup and copying is negligible so we can disable it via comment above line 97:

`//nolint:hugeParam // This function is used in setup so copying MetricsConfig is negligible.`

Also, the linter says "**consider** passing it by pointer" so it's just a hint anyway.

tpantelis (changes requested):

pkg/syncer/resource_syncer_test.go
msft-paddy14: Done. Thanks

pkg/syncer/metrics.go
msft-paddy14: Fair point. It's anyway not a public method so can't be abused as easily. Removed

pkg/syncer/metrics.go
msft-paddy14: Have changed this but kept markOwnership as pointer and not doing nil check as it's not public. since it modifies one parameter I thought we can keep both pointers there for simplicity

pkg/syncer/resource_syncer.go
```diff
@@ -145,19 +144,7 @@ func newResourceSyncer(config *ResourceSyncerConfig) (*resourceSyncer, error) {
...
-			*syncer.config.SyncCounterOpts,
-			[]string{
-				DirectionLabel,
-				OperationLabel,
-				SyncerNameLabel,
-			},
-		)
-		prometheus.MustRegister(syncer.syncCounter)
-	}
+	syncer.metrics = newSyncerMetrics(syncer.config.Metrics)
```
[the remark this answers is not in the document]
msft-paddy14: This seems highly unlikely and using same metric config by retrying on failure of syncer creation isn't a pattern. The handling for this might make for unnecessary code. Resolving it.

tpantelis (approved):

tpantelis:
@coderabbitai approve

PR #1297, opened 2026-02-26
Change convention of using slash to dot to comply with ConfigMap regex

### LINKED ISSUE #1296: ConfigMap global keys with `/` are not valid in Kubernetes clusters
**Summary:**
Keys containing `/` (forward slash) are used in multiple configuration constants in the Submariner codebase (e.g., `k8s.client/qps`, `broker-endpoint-slices.workqueue/num-workers`). These keys are loaded from the `submariner-global` ConfigMap. However, Kubernetes ConfigMap data keys do **not** allow the `/` character.

Get the following error:

` data[broker-endpoint-slices.workqueue/num-workers]: Invalid value: "broker-endpoint-slices.workqueue/num-workers": a valid config key must consist of alphanumeric characters, '-', '_' or '.' (e.g. 'key.name',  or 'KEY_NAME',  or 'key-name', regex used for validation is '[-._a-zA-Z0-9]+')`
 

**Problem:**
- Attempting to apply a ConfigMap in a real Kubernetes cluster with any key containing `/` results in an error and the resource is rejected.
- These keys appear to work in Go unit tests (which do not use the Kubernetes API server), but are not working via `kubectl`
-
- As a consequence, those global config values (burst, QPS, worker count, *etc*.) cannot actually be used in production — only keys like `use-nftables` (which contain no `/`) are honored.

**Request for Confirmation:**
- Can maintainers confirm if these keys with `/` were ever tested on a real cluster via Kubernetes API, or are they only covered in tests that construct in-memory `corev1.ConfigMap`? 
- If there is a known workaround for setting these keys, please document it. Otherwise, the implementation should be changed to use keys that conform to the allowed `[a-zA-Z0-9._-]` pattern (e.g., replace `/` with `.` or `-`).
---
If there's a workaround or I am mistaken about support for `/` in ConfigMap keys, please clarify here. Maybe we can add a shipyard test to validate this with a sample in repo?

msft-paddy14:
Replace `/` with `.` in ConfigMap global keys

Fixes https://github.com/submariner-io/admiral/issues/1296

### Problem

Several global configuration keys used in the `submariner-global` expected ConfigMap to contain `/` characters (e.g., `k8s.client/qps`, `<resource>.workqueue/num-workers`). Kubernetes ConfigMap data keys must match `[-._a-zA-Z0-9]+`, so any key with `/` is rejected by the API server

## Summary by CodeRabbit

* **Chores**
  * Updated configuration key format from slash-delimited to dot-delimited notation for Kubernetes client and broker settings and for workqueue keys, ensuring consistent global configuration key formatting across the system.

### DIFF
--- a/pkg/global/config.go
+++ b/pkg/global/config.go
@@ -33,10 +33,10 @@ type ValueType interface {
 }
 
 const (
-	K8sClientQPS         = "k8s.client/qps"
-	K8sClientBurst       = "k8s.client/burst"
-	K8sBrokerClientQPS   = "k8s.broker.client/qps"
-	K8sBrokerClientBurst = "k8s.broker.client/burst"
+	K8sClientQPS         = "k8s.client.qps"
+	K8sClientBurst       = "k8s.client.burst"
+	K8sBrokerClientQPS   = "k8s.broker.client.qps"
+	K8sBrokerClientBurst = "k8s.broker.client.burst"
 )
 
 var (
--- a/pkg/workqueue/config.go
+++ b/pkg/workqueue/config.go
@@ -83,5 +83,5 @@ func ConfigFromGlobal(keyPrefix string, defaultConfig *Config) *Config {
 }
 
 func ToConfigMapDataKey(prefix, name string) string {
-	return fmt.Sprintf("%s.workqueue/%s", prefix, name)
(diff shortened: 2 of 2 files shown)


skitt (approved):

skitt:
Thanks! We should add these to the e2e tests :wink:

PR #1312, opened 2026-03-10
Use contextual logging in util package

tpantelis:
Remove static logger from _create_or_update.go_ and use context-based logging via `logr.FromContextOrDiscard()`. This allows callers to optionally provide a logger through the context, making logging opt-in rather than globally enabled/disabled.

The previous implementation had a race condition where the static logger was initialized with `logf.Log` at package init time, then `SetMaxVerbosity(0)` was called in `init()` to disable logging. However, if the global `logf.Log` logger sink was configured after the package was initialized, the `SetMaxVerbosity(0)` call would be ineffective, causing unwanted debug logs to appear. With contextual logging, the logger is retrieved at runtime from the context, eliminating this timing issue and giving callers explicit control over logging.

Changes:
    - Remove static logger and init() function from create_or_update.go
    - Retrieve logger from context in maybeCreateOrUpdate() and createResource()
    - Preserve static helperLogger in helpers.go for package-level error hooks that cannot use context
    - Update tests to pass a context logger

## Summary by CodeRabbit

* **Refactor**
  * Improved logging to use per-operation context-aware loggers for clearer, more accurate diagnostics across resource operations.
* **Tests**
  * Test harness updated to propagate and verify context-aware logging during test runs, improving test reliability and observability.

### DIFF
--- a/pkg/util/create_or_update.go
+++ b/pkg/util/create_or_update.go
@@ -24,6 +24,7 @@ import (
 	"slices"
 	"time"
 
+	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	"github.com/submariner-io/admiral/pkg/log"
 	"github.com/submariner-io/admiral/pkg/resource"
@@ -36,7 +37,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/retry"
-	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 type OperationResult string
@@ -64,8 +64,6 @@ var backOff wait.Backoff = wait.Backoff{
 	Cap:      40 * time.Second,
 }
 
-var logger = log.Logger{Logger: logf.Log}
-
 type CreateOrUpdateOptions[T runtime.Object] struct {
 	Client         resource.Interface[T]
 	Obj            T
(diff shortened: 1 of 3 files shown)


pkg/util/create_or_update.go
```diff
@@ -253,7 +248,9 @@ func getResource[T runtime.Object](ctx context.Context, options *CreateOrUpdateO
 	return list[0], nil
 }
 
-func createResource[T runtime.Object](ctx context.Context, client resource.Interface[T], obj T, mutate MutateFn[T]) (T, error) {
+func createResource[T runtime.Object](ctx context.Context, client resource.Interface[T], obj T, mutate MutateFn[T],
```
skitt: Instead of passing the logger explicitly, why not retrieve it from the context here as well?

pkg/util/create_or_update_test.go
```diff
@@ -357,12 +361,16 @@ type createOrUpdateTestDriver struct {
...
+	loggerCtx   context.Context //nolint:containedctx // Test helper context with logger
 }
 
 func newCreateOrUpdateTestDiver() *createOrUpdateTestDriver {
 	t := &createOrUpdateTestDriver{}
 
-	BeforeEach(func() {
+	BeforeEach(func(ctx SpecContext) {
+		// Initialize context with logger for this test driver
+		t.loggerCtx = logr.NewContext(ctx, logf.Log)
```
skitt: Ginkgo [provides a new context for every node](https://github.com/onsi/ginkgo/blob/96196475f2d8b069fa8e6a4529d7938e015fc2bf/internal/suite.go#L895), and relies on that for cancellations; I don’t think this can work reliably. As far as I can determine, the tests will have to keep the logger itself in their struct, and each node will have to call `logr.NewContext(ctx, t.logger)` to get a context with a logger.

skitt (changes requested):
This is very cool — I was bothered by the direct use of the controller-runtime logger in the CoreDNS module in Lighthouse, but didn’t think of using a contextual logger!

pkg/util/create_or_update_test.go
tpantelis: yeah I was trying to avoid modifying each `It` spec to get a context logger but that's the best approach.

skitt (approved):

PR #1339, opened 2026-04-21
Switch to govulncheck for Go scanning

skitt:
govulncheck only reports vulnerabilities in reachable packages, which will greatly reduce the number of false positives.

This changes CI linting to use govulncheck, using targets provided by Shipyard.

PR checks no longer upload SARIF reports, and SARIF uploads are changed to periodic rather than per-push. This ensures that security issues are caught in a timely fashion even if no development activity occurs in the repository.

## Summary by CodeRabbit

* **Chores**
  * Moved vulnerability scanning to a periodic job that runs across maintained release branches.
  * Removed inline vulnerability scan steps from regular workflows to simplify CI runs.
  * Switched to branch-scoped vulnerability reporting with centralized uploads to ensure scan results are produced and retained per branch.

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -106,24 +106,13 @@ jobs:
   vulnerability-scan:
     name: Vulnerability Scanning
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     steps:
       - name: Check out the repository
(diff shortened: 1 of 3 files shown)


skitt:
@coderabbitai review

tpantelis (approved):

